PR #129, opened 2015-06-02
Zm client helpers

zacharym:
@paetling

### DIFF
--- a/dusty/commands/utils.py
+++ b/dusty/commands/utils.py
@@ -8,7 +8,6 @@ from ..systems.docker import get_docker_env
 def _executable_path(executable_name):
     return subprocess.check_output(['which', executable_name]).strip()
 
-
 def exec_docker(*args):
     updated_env = copy(os.environ)
     updated_env.update(get_docker_env())
--- a/dusty/commands/utils.py
+++ b/dusty/commands/utils.py
@@ -1,8 +1,16 @@
+from copy import copy
 import os
 import subprocess
+
+from ..systems.docker import get_docker_env
+
 
 def _executable_path(executable_name):
     return subprocess.check_output(['which', executable_name]).strip()
 
+
 def exec_docker(*args):
-    os.execl(_executable_path('docker'), 'docker', *args)
+    updated_env = copy(os.environ)
+    updated_env.update(get_docker_env())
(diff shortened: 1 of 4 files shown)


dusty/commands/utils.py
```diff
@@ -1,8 +1,16 @@
...
 import os
 import subprocess
 
+from ..systems.docker import get_docker_env
+
+
 def _executable_path(executable_name):
     return subprocess.check_output(['which', executable_name]).strip()
 
+
```
paetling: remove this space

dusty/commands/utils.py
```diff
@@ -1,8 +1,16 @@
...
 def exec_docker(*args):
-    os.execl(_executable_path('docker'), 'docker', *args)
+    updated_env = copy(os.environ)
+    updated_env.update(get_docker_env())
+    args = args + (updated_env,)
+    os.execle(_executable_path('docker'), 'docker', *args)
```
paetling: how does this work?  You are adding environment variables as arguments to docker.

dusty/commands/utils.py
zacharym: Ok. My linter complains at me for only having one, but I suppose it's better to be consistent throughout the project

dusty/commands/utils.py
zacharym: it used to be an `os.execl` command which runs the executable and causes the new process to inherit the environment of the of the current process. `os.execle ` tells it that we will be providing the environment. Original I had just provided the docker_envs, but I wanted to preserve current behavior but augment it with the docker_env vars

zacharym: the issue is the way we are using *args because os.execle expects the env dict to be the last argument. So I think it's legit to do it this way

zacharym: since excele depends that an env be provided I think it parses that arg first, the reads the executable and arguments for that

zacharym: https://docs.python.org/2/library/os.html#os.execle

paetling:
Awesome.  You have tried this and it works?

zacharym:
yep

paetling:
:ghost: :cow:

PR #132, opened 2015-06-02
replace keys in config with constants

paetling:
review @thieman

### DIFF
--- a/dusty/cli/__init__.py
+++ b/dusty/cli/__init__.py
@@ -28,14 +28,14 @@ import socket
 
 from docopt import docopt
 
-from ..constants import SOCKET_PATH, SOCKET_TERMINATOR, SOCKET_ERROR_TERMINATOR
 from ..config import get_config_value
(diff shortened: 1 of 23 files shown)


dusty/cli/__init__.py
```diff
@@ -28,14 +28,14 @@
 
 from docopt import docopt
 
-from ..constants import SOCKET_PATH, SOCKET_TERMINATOR, SOCKET_ERROR_TERMINATOR
 from ..config import get_config_value
 from ..log import configure_client_logging, log_to_client
 from ..payload import Payload
 from . import bundles, config, cp, dump, disk, logs, repos, restart, script, shell, stop, sync, up, validate, setup
+from dusty import constants
```
thieman: Relative imports only

paetling: will change

thieman:
Is this necessary? If we're going to do this can we at least make sure we get something out of it, e.g. will pylint fail our tests if we reference a constant that doesn't exist?

PR #135, opened 2015-06-03
Ae scripts arguments

paetling:
review @jsingle

### DIFF
--- a/dusty/commands/bundles.py
+++ b/dusty/commands/bundles.py
@@ -14,6 +14,9 @@ def list_bundles():
         table.add_row([bundle,
                        bundle_spec['description'],
                        u"✓" if bundle in activated_bundles else ""])
+    import logging
+    logging.error(u"✓")
+    logging.error(table.get_string())
     log_to_client(table.get_string(sortby="Name"))
 
 def activate_bundle(bundle_names):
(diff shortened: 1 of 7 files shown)


dusty/commands/bundles.py
```diff
@@ -14,6 +14,9 @@ def list_bundles():
         table.add_row([bundle,
                        bundle_spec['description'],
                        u"✓" if bundle in activated_bundles else ""])
+    import logging
```
jsingle: rogue import/debug code

dusty/commands/script.py
```diff
@@ -1,5 +1,6 @@
-import textwrap
+# coding=utf-8
```
jsingle: What's this?

jsingle:
Couple minor things but this looks good.  Did you consider always passing any additional arguments to the script, rather than requiring a key in the specs to be set?

dusty/commands/script.py
paetling: this encodes the python file in unicode, which allows the check mark to be handled properly

paetling:
I did consider it.  But i figure if you are adding a script to your .yml, you know at that time if it should take arguemnts.  If it should not, you really do not want people passing arguments to it.

PR #137, opened 2015-06-03
added a plist file!!

jsingle:
@paetling this actually took me a while to make/debug.  Can it just live in the dusty repo here?


org.gc.dusty.plist
```diff
@@ -0,0 +1,29 @@
...
+  </dict>
+  <key>KeepAlive</key>
+  <true/>
+  <key>Label</key>
+  <string>org.gc.dustyd</string>
+  <key>ProgramArguments</key>
+  <array>
+    <string>bash</string>
+    <string>-c</string>
+    <string>export PATH=$PATH:/usr/local/bin; env; dustyd</string>
```
paetling: why do you need the env command?

paetling:
I think it is fine to stay if we create a folder like `basic_setup` or `example_setup`.  Then added in a readme in that file or what these things were.   Because this is a very gc specific thing (we want to run it like this) but it could be useful to people down the road

org.gc.dusty.plist
jsingle: oops that's debugging.  The path mod is needed because OSX daemons/agents seem to use a really bare-bones environment, which doesn't have `PATH` set normally

jsingle:
I'm unsure exactly how gc-specific it is... travis seems to think that making a service to run dustyd automatically is what everyone would want.  I'll leave the PR open for now

paetling:
ok

PR #139, opened 2015-06-03
add in checking of our dusty specs in dusty repo

paetling:
review @jsingle


jsingle:
Looks good if the tests pass :boom:

thieman:
@paetling Isn't this going to make it impossible to ship backwards-incompatible changes to the specs? You try to change Dusty, but it fails because the old specs don't validate against new Dusty, so then you try to change the specs, but that fails because the new specs don't validate against old Dusty...

PR #141, opened 2015-06-03
rename warnings file to dusty_warnings

zacharym:
@jsingle


jsingle:
:camel: :cactus:

thieman:
@zacharym As discussed, I'm in favor of reverting this. The stdlib warnings module can be imported by using `from __future__ import absolute_import`, which backports the much-less-ambiguous import behavior of Python 3.

PR #142, opened 2015-06-03
Js install readme steps

jsingle:
@paetling these are currently the install steps (for any user).  For gc users, should the script go in gcsystems (it could do this + other gc-specific stuff)?


paetling:
this looks good.  what are the pros and cons of doing it in gcsystems vs gcconfig?  Not really sure what gcconfig is supposed to be used for

jsingle:
Either way, I don't really know what either is for

paetling:
just put it in gcsystems.  you can move it later if needed

PR #144, opened 2015-06-03
.get is nice

jsingle:
@paetling cc @pdv

### DIFF
--- a/dusty/systems/rsync/__init__.py
+++ b/dusty/systems/rsync/__init__.py
@@ -60,7 +60,7 @@ def sync_repos_by_app_name(app_names):
     repos = set()
     assembled_specs = get_assembled_specs()
     for app_name in app_names:
-        for lib_name in assembled_specs['apps'][app_name]['depends']['libs']:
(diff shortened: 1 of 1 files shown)


dusty/systems/rsync/__init__.py
```diff
@@ -60,7 +60,7 @@ def sync_repos_by_app_name(app_names):
     repos = set()
     assembled_specs = get_assembled_specs()
     for app_name in app_names:
-        for lib_name in assembled_specs['apps'][app_name]['depends']['libs']:
+        for lib_name in assembled_specs['apps'][app_name].get('depends', {}).get('libs', []):
```
paetling: is depends optional? / arguments inside it optional?

paetling:
just looked. :ghost: :cow:

dusty/systems/rsync/__init__.py
jsingle: I'd say that both are optional, yes.  An app doesn't need to have dependencies, and it definitely does not need to depend on libs (allegro doesn't, that's what caused this error).

PR #147, opened 2015-06-04
Add magic comment to stop our CI from building a package

thieman:
@jsingle


jsingle:
Haha is this a normal way to do this?

PR #148, opened 2015-06-04
Zm dustyd suppress warnings

zacharym:
@jsingle cc @thieman


jsingle:
Looks good assuming you tested it with & without the flag

zacharym:
@thieman updated per your recommendations

PR #149, opened 2015-06-04
Add config integration tests

thieman:
@paetling Config integration tests with some related changes:

1. Client now strips socket terminators out of its output. This was preventing us from converting stdout to YAML when appropriate for validation.
1. Added ability in the integration test case to only return stdout returned past a specified point
1. Standard integration test setUp should now prevent daemon warnings being shown
1. run_command now raises if the client returns a non-0 exit code


tests/testcases.py
```diff
@@ -67,20 +71,42 @@ def setUp(self):
...
+        # Run this through the daemon itself so it refreshes its view of daemon warnings
+        self.run_command('config set {} {}'.format(constants.CONFIG_MAC_USERNAME_KEY, self.current_user))
         override_repo(get_specs_repo(), self.overridden_specs_path)
+        self.clear_stdout()
 
     def tearDown(self):
         shutil.rmtree(self.overridden_specs_path)
 
+    def clear_stdout(self):
+        self.stdout_start = len(sys.stdout.getvalue())
```
paetling: so we want this clearing feature so you can do multiple commands and only get the output for some of them in your checks for what is in lines?   So i think it is a good idea to be able to isolate these, but I would favor not forcing the test writer to remember to do this.   Why not, on run_command, get the stdout then, save it in a variable. Then when you check what is in stdout, use that variable.  That way as you go through you have access to that output until you mkae another command, and then you only have access to that output.    I can envision (and have made similar mistakes in the past) where you do it with a clear, pepole forget and then spend an hour debugging why there test is not working

thieman: smart

PR #154, opened 2015-06-04
take out the validate step

paetling:
review @thieman


paetling:
already did the downstream stuff

PR #157, opened 2015-06-05
add a second name config test

zacharym:
@thieman this might be stupid, but I realized we really need to have two of these to ensure that it actually works in the unfortunate case that the user's name actually is steve..


zacharym:
nvm. I am tired.

thieman:
Yeah generally if your test actually relies on the thing not existing you're supposed to use a more outlandish value, like `/this/dir/for/sure/does/not/exist` or `definitely-not-a-real-username`

PR #158, opened 2015-06-05
Convert SSH raise to a warning

thieman:
@jsingle This was making it impossible to run tests :cry:


jsingle:
OK this should be fine too

PR #159, opened 2015-06-05
Massive refactor of Repo management

thieman:
@paetling The various functions for managing Dusty's idea of what a "repo" is were getting out of hand. I've consolidated all of that functionality into a class. This broke basically the entire app.

Plan:
1. Get this massive thing merged and hopefully don't break the entire app
1. Write some tests for the Repo class
1. Usability / consistency improvements around how things are calling this. We have a mix of "passing the repo path" and "passing the Repo object" right now.

### DIFF
--- a/dusty/compiler/spec_assembler.py
+++ b/dusty/compiler/spec_assembler.py
@@ -4,9 +4,8 @@ import yaml
 import logging
 
 from ..config import get_config_value
-from ..path import local_repo_path
 from .. import constants
-
+from ..source import Repo
 
 def _get_dependent(dependent_type, name, specs, root_spec_type):
     """
@@ -105,10 +104,10 @@ def get_assembled_specs():
     return specs
 
 def get_specs_repo():
-    return get_config_value(constants.CONFIG_SPECS_REPO_KEY)
+    return Repo(get_config_value(constants.CONFIG_SPECS_REPO_KEY))
 
 def get_specs_path():
-    return local_repo_path(get_specs_repo())
+    return get_specs_repo().local_path
 
 def get_specs():
     specs_path = get_specs_path()
@@ -143,5 +142,5 @@ def get_all_repos(active_only=False, include_specs_repo=True):
     for type_key in ['apps', 'libs']:
         for spec in specs[type_key].itervalues():
             if 'repo' in spec:
-                repos.add(spec['repo'])
+                repos.add(Repo(spec['repo']))
     return repos
--- a/dusty/source.py
+++ b/dusty/source.py
@@ -11,36 +11,7 @@ from .config import get_config_value
 from . import constants
 from .notifier import notify
 from .log import log_to_client
-from .compiler.spec_assembler import get_all_repos
-from .path import local_repo_path, managed_repo_path, parent_dir
-
-def repo_is_overridden(repo_name):
-    return repo_name in get_config_value(constants.CONFIG_REPO_OVERRIDES_KEY)
-
-
-def short_repo_name(repo_name):
-    return repo_name.split('/')[-1]
-
-def _is_short_name(repo_name):
-    return '/' not in repo_name
-
-def _expand_repo_name(passed_short_name):
-    match = None
-    for repo_name in get_all_repos():
-        short_name = short_repo_name(repo_name)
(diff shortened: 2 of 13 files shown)


nrschultz:
good prioritization, all #1.

dusty/compiler/spec_assembler.py
```diff
@@ -105,10 +104,10 @@ def get_assembled_specs():
     return specs
 
 def get_specs_repo():
-    return get_config_value(constants.CONFIG_SPECS_REPO_KEY)
+    return Repo(get_config_value(constants.CONFIG_SPECS_REPO_KEY))
 
 def get_specs_path():
-    return local_repo_path(get_specs_repo())
+    return get_specs_repo().local_path
```
paetling: shouldnt this just be called something like path.  Which rerturns local if overridden and managed if not

dusty/source.py
```diff
@@ -52,32 +23,93 @@ def git_error_handling():
...
+        return self.remote_path in get_config_value(constants.CONFIG_REPO_OVERRIDES_KEY)
+
+    @property
+    def override_path(self):
+        repo_overrides = get_config_value(constants.CONFIG_REPO_OVERRIDES_KEY)
+        return repo_overrides.get(self.remote_path)
+
+    @property
+    def local_path(self):
+        return self.override_path if self.is_overridden else self.managed_path
```
paetling: this names implies to me that path on my local machine (opposite of the remote path).  But it actually means either the remote path or the managed path depending on if i am overriding.    path_on_machine?

tests/testcases.py
```diff
@@ -33,7 +33,7 @@ def setUp(self):
         constants.CONFIG_PATH = self.temp_config_path
         write_default_config()
         save_config_value(constants.CONFIG_SPECS_REPO_KEY, 'github.com/org/dusty-specs')
-        override_repo(get_specs_repo(), self.temp_specs_path)
+        override_repo(get_specs_repo().remote_path, self.temp_specs_path)
```
paetling: other then circular import problem, why is this not a instance method on Repo?

paetling:
Also, Where are we passing repo objects around in the old code?

dusty/compiler/spec_assembler.py
paetling: if you move these two functions to path.py (arguable one of them should live there anyway) and get_all_repos.py to source.py.  The circular import goes away.  resolve and override then become easier/simpler

thieman: get_all_repos is based on the specs, so the "proper" place for it is probably in the spec assembler

tests/testcases.py
thieman: Don't understand this comment

thieman:
Repo objects don't exist in the old code, I changed some places to pass around objects but we are definitely still passing strings around in other places. We should standardize on objects.

thieman:
@nrschultz MARKDOWN HAS FAILED ME

dusty/source.py
thieman: it never means the remote path, it is either the local path of your override directory or the local path of the managed directory

paetling: i know that. just saying that it sounds like it always means override directory

tests/testcases.py
paetling: i think override repo should be an instance method.  it seems like the only reason it is not, is so that there are not a circular import problem?

paetling:
You said in your pr comment that in the old code (prior to this pr) we were passing around repo objects: Usability / consistency improvements around how things are calling this. We have a mix of "passing the repo path" and "passing the Repo object" right now.  I just do not see where that is happening?

tests/testcases.py
thieman: Oh `override_repo` you mean? It has a lot of app logic in it (config mutation, log to client, etc.) that would pollute the Repo class.

thieman: Though this will eventually take a `Repo` object rather than a path.

thieman:
"right now" meant "once this PR is merged"

paetling:
ohh.  ok.

paetling:
pr seems fine. like the standardization

PR #160, opened 2015-06-05
Js general install script

jsingle:
@paetling

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Docker-based development environment manager
 Currently the recommended way to install dusty is to use our installation script, which will install the latest
 version of dusty's binaries, and set up a dustyd daemon plist:
 ```
-bash `curl -L https://raw.githubusercontent.com/gamechanger/dusty/master/setup/install.sh`
+bash -c "`curl -L https://raw.githubusercontent.com/gamechanger/dusty/master/setup/install.sh`"
 ```
 
 This will:
--- a/README.md
+++ b/README.md
@@ -4,22 +4,12 @@ Docker-based development environment manager
 
 
 # Installation
-Currently the recommended way to install dusty is to clone the dusty repository, create dusty binaries using our script, and place them in your `/usr/local/bin` folder.
-These commands should do that:
+Currently the recommended way to install dusty is to use our installation script, which will install the latest
+version of dusty's binaries, and set up a dustyd daemon plist:
 ```
-git clone https://github.com/gamechanger/dusty.git
-cd dusty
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -4,22 +4,17 @@ Docker-based development environment manager
...
-These commands should do that:
+Currently the recommended way to install dusty is to use our installation script, which will install the latest
+version of dusty's binaries, and set up a dustyd daemon plist:
 ```
-git clone https://github.com/gamechanger/dusty.git
-cd dusty
-./setup/create_binaries.sh
-cp dist/dusty /usr/local/bin/dusty
-cp dist/dustyd /usr/local/bin/dustyd
+bash `curl -L https://raw.githubusercontent.com/gamechanger/dusty/master/setup/install.sh`
```
paetling: we are not versioning our install script?  just sending them to our most recent one in github?

jsingle: You're right currently we're only versioning binaries, not the install script or the plist, which are also curled.  We should probably version those

setup/install.sh
```diff
@@ -0,0 +1,10 @@
+release=0.0.1
+echo "Downloading dusty files"
+curl -L https://github.com/gamechanger/dusty/releases/download/$release/dusty > /usr/local/bin/dusty
+curl -L https://github.com/gamechanger/dusty/releases/download/$release/dustyd > /usr/local/bin/dustyd
```
paetling: these two lines work?  Just the binary is being place in these locations?

jsingle: Yeah I think it works... why wouldn't it?

setup/install.sh
```diff
@@ -0,0 +1,10 @@
...
+echo "Authenticating as super user... needed to setup daemon"
+sudo -v
+sudo curl -L -o /System/Library/LaunchDaemons/org.gamechanger.dustyd.plist https://raw.githubusercontent.com/gamechanger/dusty/master/setup/org.gamechanger.dustyd.plist
```
paetling: why do you do the outputting to a file one way above and another way here?

setup/install.sh
jsingle: Hmm maybe using a curl flag instead of output redirection would be better?

setup/install.sh
jsingle: because when you do `sudo` I think you'd get permission denied on the output redirection command.  The curl has superuser permission, but that doesn't change the output redirection permissions

paetling:
seems legit.

PR #163, opened 2015-06-05
Js preflight only

jsingle:
@thieman

### DIFF
--- a/dusty/preflight.py
+++ b/dusty/preflight.py
@@ -89,10 +89,8 @@ def _ensure_config_dir_exists():
 
 def preflight_check():
     logging.info('Starting preflight check')
-    errors = []
-    for check in [_check_nginx, _check_rsync, _check_virtualbox, _check_boot2docker, _check_docker,
-                  _check_docker_compose, _assert_hosts_file_is_writable]:
-        errors.append(check)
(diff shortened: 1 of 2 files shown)


dusty/preflight.py
```diff
@@ -16,6 +16,15 @@
...
     pass
 
+def returns_exception(f):
+    def inner():
+        try:
+            f()
+            return None
+        except Exception as e:
+            return e
+    return inner
```
thieman: `if err != nil`

dusty/preflight.py
```diff
@@ -73,14 +89,18 @@ def _ensure_config_dir_exists():
...
 def preflight_check():
     logging.info('Starting preflight check')
+    errors = []
+    errors.append(_check_nginx())
+    errors.append(_check_rsync())
+    errors.append(_check_virtualbox())
+    errors.append(_check_boot2docker())
+    errors.append(_check_docker())
+    errors.append(_check_docker_compose())
+    errors.append(_assert_hosts_file_is_writable())
```
thieman: Get thunky with it

```
for check in [_check_nginx, ...]:
    errors.append(check())
```

You can take advantage of them all having the same call signature.

thieman: Shit you can even just do a list comprehension like a boss

jsingle: sounds better

thieman:
Cool PR bro

PR #164, opened 2015-06-05
Add ability to clone local remotes

thieman:
@paetling Maybe if this works I can actually get back to integration testing! Hooray!

### DIFF
--- a/dusty/source.py
+++ b/dusty/source.py
@@ -100,7 +100,10 @@ class Repo(object):
         if not os.path.exists(repo_path_parent):
             os.makedirs(repo_path_parent)
         with git_error_handling():
-            managed_repo = git.Repo.clone_from('ssh://{}@{}'.format(constants.GIT_USER, self.remote_path), self.managed_path)
+            if self.is_local_repo:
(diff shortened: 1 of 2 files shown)


dusty/source.py
```diff
@@ -100,7 +100,10 @@ def ensure_local_repo(self):
         if not os.path.exists(repo_path_parent):
             os.makedirs(repo_path_parent)
         with git_error_handling():
-            managed_repo = git.Repo.clone_from('ssh://{}@{}'.format(constants.GIT_USER, self.remote_path), self.managed_path)
+            if self.is_local_repo:
+                git.Repo.clone_from('file:///{}'.format(self.remote_path), self.managed_path)
```
paetling: why are we doing the cloning in this case?

paetling: is this if they setup the repo for this to be /gc/gcsystems?

paetling: seems like a weird pattern.  but i guess it would work.  It would continually pull from a local git repo to the dusty managed version.  Interesting way to do local dev without overrides needed

PR #169, opened 2015-06-06
refactor some of the docker-compose commands and the saving of the docker file

paetling:
review @thieman     not sure there is much else that can be done here.   I think all of the compiler stuff is pretty specific and cannot really be broken out.  Planning on just making a new compiler file for tests.

### DIFF
--- a/dusty/systems/docker/compose.py
+++ b/dusty/systems/docker/compose.py
@@ -8,16 +8,17 @@ from ... import constants
 from ...log import log_to_client
 from ...subprocess import check_output_demoted, check_and_log_output_and_error_demoted
 from ...compiler.spec_assembler import get_expected_number_of_running_containers
+from ...path import parent_dir
 
 def _write_composefile(compose_config, compose_file_location):
     logging.info('Writing new Composefile')
-    compose_dir_location = '/'.join(compose_file_location.split('/')[0:-1])
+    compose_dir_location = parent_dir(compose_file_location)
     if not os.path.exists(compose_dir_location):
         os.makedirs(compose_dir_location)
     with open(compose_file_location, 'w') as f:
         f.write(yaml.dump(compose_config, default_flow_style=False))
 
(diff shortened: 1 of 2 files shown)


dusty/systems/docker/compose.py
```diff
@@ -9,25 +9,35 @@
 from ...subprocess import check_output_demoted, check_and_log_output_and_error_demoted
 from ...compiler.spec_assembler import get_expected_number_of_running_containers
 
-def _write_composefile(compose_config):
+def _write_composefile(compose_config, compose_file_location):
     logging.info('Writing new Composefile')
-    if not os.path.exists(constants.COMPOSE_DIR):
-        os.makedirs(constants.COMPOSE_DIR)
-    with open(constants.COMPOSEFILE_PATH, 'w') as f:
+    compose_dir_location = '/'.join(compose_file_location.split('/')[0:-1])
```
thieman: Think `os.path.split` does what you're trying to do here

thieman: Or the `parent_dir` helper we've already got in `path`

dusty/systems/docker/compose.py
```diff
@@ -9,25 +9,35 @@
...
 
-def _compose_up(recreate_containers=True):
+def _compose_up(recreate_containers=True, compose_file_location=None, project_name=None):
     logging.info('Running docker-compose up')
-    command = ['docker-compose', '-f', constants.COMPOSEFILE_PATH, '-p', 'dusty', 'up', '-d', '--allow-insecure-ssl']
+    command = ['docker-compose']
+    if compose_file_location is not None:
+        command += ['-f', compose_file_location]
+    if project_name is not None:
+        command += ['-p', project_name]
```
thieman: These two should probably just be required, yeah?

thieman:
Good changes :bear:

PR #173, opened 2015-06-07
Refactoring of compose compiler and spec_assembler to make testing stuff easier

paetling:
review @thieman


thieman:
This seems fine. I still agree with the idea that the specs stuff should probably be wrapped in a class at this point.

PR #176, opened 2015-06-08
makes follow flag not default for logs tail, adds some tests

zacharym:
@paetling cc @thieman

### DIFF
--- a/dusty/cli/logs.py
+++ b/dusty/cli/logs.py
@@ -2,17 +2,17 @@
 or service.
 
 Usage:
-  logs [--tail=NUM] <service>
+  logs [-f] [--tail=NUM] <service>
 
 Options:
+  -f          follow log output
   --tail=NUM  show NUM lines from end of file
 
 """
-
 from docopt import docopt
 
 from ..commands.logs import tail_container_logs
 
 def main(argv):
     args = docopt(__doc__, argv)
-    return tail_container_logs(args['<service>'], args['--tail'])
+    return tail_container_logs(args['<service>'], args['-f'], args['--tail'])
--- a/dusty/commands/logs.py
+++ b/dusty/commands/logs.py
@@ -1,9 +1,12 @@
 from ..systems.docker import get_dusty_containers
 from .utils import exec_docker
 
-def tail_container_logs(app_or_service_name, lines=None):
+def tail_container_logs(app_or_service_name, follow=False, lines=None):
     container = get_dusty_containers([app_or_service_name])[0]
-    if lines is None:
-        exec_docker('logs', '-f', container['Id'])
-    else:
-        exec_docker('logs', '-f', '--tail={}'.format(lines), container['Id'])
+    args = ['logs']
+    if follow:
+        args.append('-f')
+    if lines:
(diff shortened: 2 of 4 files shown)


paetling:
update your tests.  taking a look

dusty/cli/logs.py
```diff
@@ -15,4 +17,4 @@
 
 def main(argv):
     args = docopt(__doc__, argv)
-    return tail_container_logs(args['<service>'], args['--tail'])
+    return tail_container_logs(args['<service>'], args['-f'], args['--tail'])
```
paetling: so i think @thieman moved away from having just normal tail be the default.  I think it is fine to change back, but might want to consult him

dusty/commands/logs.py
```diff
@@ -1,9 +1,12 @@
...
 from .utils import exec_docker
 
-def tail_container_logs(app_or_service_name, lines=None):
+def tail_container_logs(app_or_service_name, follow=False, lines=None):
     container = get_dusty_containers([app_or_service_name])[0]
-    if lines is None:
-        exec_docker('logs', '-f', container['Id'])
-    else:
-        exec_docker('logs', '-f', '--tail={}'.format(lines), container['Id'])
+    args = ('logs',)
```
paetling: why not just make it a list. instead of making it tuple like this

dusty/commands/logs.py
```diff
@@ -1,9 +1,12 @@
...
+    if follow:
+        args += ('-f',)
+    if lines:
+        args += ('--tail={}'.format(lines),)
+    args += (container['Id'],)
+    exec_docker(args)
```
paetling: also i think you will need *args

dusty/commands/utils.py
```diff
@@ -8,8 +8,8 @@
 def _executable_path(executable_name):
     return subprocess.check_output(['which', executable_name]).strip()
 
-def exec_docker(*args):
+def exec_docker(args):
```
paetling: you probably do not want to change exec_docker

dusty/cli/logs.py
```diff
@@ -2,12 +2,14 @@
...
 Usage:
-  logs [--tail=NUM] <service>
+  logs [-f] [--tail=NUM] <service>
 
 Options:
+  -f          follow log output
   --tail=NUM  show NUM lines from end of file
 
 """
+import sys
```
thieman: extraneous

dusty/commands/logs.py
thieman: Agreed, much more idiomatic to use lists for this sort of things. Tuples are immutable. This `+=` assignment technically works but it's very strange.

dusty/commands/utils.py
thieman: What was the motivation for this?

dusty/commands/logs.py
zacharym: I didn't realize that passing some *args would cast it as a tuple for you. nice

zacharym:
@paetling @thieman this is updated

paetling:
:ghost: :cow: seems good to me

PR #177, opened 2015-06-09
Kill notify

thieman:
@paetling I thought this would be necessary back when we were planning on having the auto-reloading feature. It's not now, and it's a shitty library, so let's kill it.


paetling:
:ghost: :cow:

PR #178, opened 2015-06-09
catch update repo fail and still finish up

zacharym:
@paetling @thieman

### DIFF
--- a/dusty/source.py
+++ b/dusty/source.py
@@ -112,4 +112,8 @@ class Repo(object):
 
         managed_repo = git.Repo(self.managed_path)
         with git_error_handling():
-            managed_repo.remote().pull('master')
+            try:
+                managed_repo.remote().pull('master')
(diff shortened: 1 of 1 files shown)


dusty/source.py
```diff
@@ -114,5 +114,8 @@ def update_local_repo(self):
         notify('Pulling latest updates for {}'.format(self.short_name))
 
         managed_repo = git.Repo(self.managed_path)
-        with git_error_handling():
+        try:
             managed_repo.remote().pull('master')
+        except git.exc.GitCommandError as e:
+            log_to_client('Could not update dusty managed remote: {}'.format(managed_repo.remote().url))
+            log_to_client('Failed with error: {}'.format(e))
```
thieman: Why not just add a raise flag to the context manager and keep using that? Now they'll never get the helpful message about ssh agent.

zacharym:
going to add a no-pull flag to up instead

PR #179, opened 2015-06-09
Kill our fancy checkmarks

thieman:
@paetling Default sysout encoding in Python 2 is ASCII, we should stop fighting against that


paetling:
its a sad day for dusty  :( :cow:

PR #186, opened 2015-06-10
Support pre-Yosemite launchd configurations in SSH config

thieman:
@zacharym cc @jsingle crazy shit

### DIFF
--- a/dusty/config.py
+++ b/dusty/config.py
@@ -9,6 +9,9 @@ import os
 import pwd
 import subprocess
 import yaml
+import platform
+
+import psutil
 
 from . import constants
 from .warnings import daemon_warnings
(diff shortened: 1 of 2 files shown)


dusty/config.py
```diff
@@ -61,6 +64,45 @@ def verify_mac_username(username):
...
+    _set_ssh_auth_sock(ssh_auth_sock)
+
+def _load_ssh_auth_pre_yosemite():
+    """For OS X versions before Yosemite, many launchd processes run simultaneously under
+    different users and different permission models. The simpler `asuser` trick we use
+    in Yosemite doesn't work, since it gets routed to the wrong launchd. We instead need
+    to find the running ssh-agent process and use its PID to navigate ourselves
+    to the correct launchd."""
+    for process in psutil.process_iter():
+        if process.name() == 'ssh-agent':
```
zacharym: are we sure that only one ssh-agent process will run at a time?

zacharym:
:+1:

dusty/config.py
thieman: No, which is why I designed it to try all the agents it can find and stop if it finds one that gives it a valid listener

PR #189, opened 2015-06-10
We are an app, we should not use floating versions

thieman:
cc @paetling


paetling:
other than you broke all the things. seems legit

PR #190, opened 2015-06-10
Ae use new specs class/refactor

paetling:
review @thieman

### DIFF
--- a/dusty/commands/script.py
+++ b/dusty/commands/script.py
@@ -10,7 +10,7 @@ def script_info_for_app(app_name):
     app_specs = get_specs()['apps'].get(app_name)
     if not app_specs:
         raise KeyError('No app found named {} in specs'.format(app_name))
-    if len(app_specs['scripts']) == 0:
+    if not app_specs['scripts']:
         log_to_client('No scripts registered for app {}'.format(app_name))
         return
 
--- a/dusty/commands/script.py
+++ b/dusty/commands/script.py
@@ -10,7 +10,7 @@ def script_info_for_app(app_name):
     app_specs = get_specs()['apps'].get(app_name)
     if not app_specs:
         raise KeyError('No app found named {} in specs'.format(app_name))
-    if len(app_specs['scripts']) > 0:
+    if len(app_specs['scripts']) == 0:
         log_to_client('No scripts registered for app {}'.format(app_name))
         return
 
@@ -25,7 +25,7 @@ def execute_script(app_name, script_name, script_arguments=[]):
     if not app_specs:
         raise KeyError('No app found named {} in specs'.format(app_name))
     script_spec = None
-    for script_dict in app_specs['scripts'], []):
+    for script_dict in app_specs['scripts']:
         if script_dict['name'] == script_name:
             script_spec = script_dict
     if script_spec is None:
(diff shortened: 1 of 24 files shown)


thieman:
image

Paddin dem stats

dusty/commands/script.py
```diff
@@ -10,14 +10,14 @@ def script_info_for_app(app_name):
     app_specs = get_specs()['apps'].get(app_name)
     if not app_specs:
         raise KeyError('No app found named {} in specs'.format(app_name))
-    if not app_specs.get('scripts'):
+    if len(app_specs['scripts']) == 0:
```
thieman: Why?

dusty/compiler/spec_assembler.py
```diff
@@ -140,15 +144,27 @@ def get_repo_of_app_or_library(app_or_library_name):
         return Repo(specs['libs'][app_or_library_name]['repo'])
     raise KeyError('did not find app or service with name {}'.format(app_or_library_name))
 
+def _get_respective_schema(specs_type):
+    if specs_type == 'apps':
+        return app_schema
+    elif specs_type == constants.CONFIG_BUNDLES_KEY:
```
thieman: Seems wrong to use the constant here

dusty/compiler/spec_assembler.py
```diff
@@ -140,15 +144,27 @@ def get_repo_of_app_or_library(app_or_library_name):
...
+        return bundle_schema
+    elif specs_type == 'libs':
+        return lib_schema
```
thieman: Guessing this is supposed to return `None` if you give it `services`. Should probably raise if you give it anything else, the bit a few lines down where you dispatch on the truthiness of this seems a bit risky otherwise.

tests/unit/utils.py
```diff
@@ -0,0 +1,20 @@
+from dusty.schemas.base_schema_class import DustySchema
+from dusty.schemas.app_schema import app_schema
+from dusty.schemas.lib_schema import lib_schema
+from dusty.schemas.bundle_schema import bundle_schema
+
+
+def get_app_dusty_schema(doc):
+    if 'image' not in doc and 'build' not in doc:
```
thieman: Why is this a legitimate case?

tests/unit/utils.py
```diff
@@ -0,0 +1,20 @@
...
+        doc['image'] = ''
+    if 'repo' not in doc:
+        doc['repo'] = ''
+    return DustySchema(app_schema, doc)
+
+def get_lib_dusty_schema(doc):
+    if 'repo' not in doc:
+        docp['repo'] = ''
```
thieman: Typo

thieman:
Minor shit otherwise :bear:

dusty/commands/script.py
paetling: we default scripts to a list now.  But you might be asking why I am not just checking that it is an empty list.  That is a good point

dusty/compiler/spec_assembler.py
paetling: I agree. was following a dumb pattern will fix.

dusty/compiler/spec_assembler.py
paetling: That is a good point. will do that.

dusty/commands/script.py
thieman: Yeah sorry that's what I was referencing

tests/unit/utils.py
paetling: This is in testing utils.  The idea is that I want to shelter those writing tests from needing to remeber all of the required fields.  So if they do not pass them in, i provide a default.

PR #194, opened 2015-06-10
Add CLI integration tests for disk commands

thieman:
@zacharym


thieman:
i'ma merge this so i can use these new test helpers in the new stuff i'm writing

PR #197, opened 2015-06-10
Allow viewing logs of dead containers

thieman:
@paetling This functionality will be tested through integration tests


paetling:
ha, we already had this functionality just had to expose it.   Your failing tests, but code looks fine

PR #201, opened 2015-06-12
Stream test results to client

thieman:
cc @zacharym also passes the exit code of the test run to the `dusty test` process

### DIFF
--- a/dusty/commands/test.py
+++ b/dusty/commands/test.py
@@ -37,7 +37,8 @@ def _construct_test_command(spec, suite_name, test_arguments):
             break
     if suite_command is None:
         raise RuntimeError('{} is not a valid suite name'.format(suite_name))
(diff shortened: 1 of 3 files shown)


dusty/commands/test.py
```diff
@@ -36,7 +37,8 @@ def _construct_test_command(spec, suite_name, test_arguments):
             break
     if suite_command is None:
         raise RuntimeError('{} is not a valid suite name'.format(suite_name))
-    test_command = '{} {}'.format(suite_command, ' '.join(test_arguments))
+    sub_command = "{} {}".format(suite_command, ' '.join(test_arguments))
+    test_command = 'sh -c "{}"'.format(sub_command.strip())
```
paetling: why do we need this tom foolery with the sh -c instead of running directly?

thieman: I think I need this to chain together commands with `&&`, don't I?

PR #202, opened 2015-06-13
Refactoring pass on test command

thieman:
WIP

### DIFF
--- a/dusty/commands/test.py
+++ b/dusty/commands/test.py
@@ -4,7 +4,7 @@ import sys
 from .. import constants
 from ..compiler.spec_assembler import get_expanded_libs_specs
 from ..compiler.compose import get_volume_mounts, get_testing_compose_dict
-from ..systems.docker.testing_image import ensure_image_exists, test_image_name
+from ..systems.docker.testing_image import ensure_test_image, test_image_name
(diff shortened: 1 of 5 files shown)


thieman:
cc @paetling basic objective here was to reduce the amount of parameters flying around. Could still use more work on docstrings and that sort of thing but I think after this it's clean enough that we should be able to build some more functionality on top without major pains.

dusty/commands/test.py
```diff
@@ -3,31 +3,33 @@
...
 from ..systems.docker import get_docker_client
 from ..systems.docker.compose import write_composefile, compose_up
 from ..systems.rsync import sync_repos_by_app_name, sync_repos_by_lib_name
 from ..log import log_to_client
 
 def run_app_or_lib_tests(app_or_lib_name, suite_name, test_arguments, force_recreate=False):
-    docker_client = get_docker_client()
+    client = get_docker_client()
     expanded_specs = get_expanded_libs_specs()
+    volumes = get_volume_mounts(app_or_lib_name, expanded_specs)
```
paetling: if you are not passing this, it is not needed

thieman: good spot

PR #203, opened 2015-06-14
Allow default_args in test spec, also add test schema

thieman:
cc @paetling

### DIFF
--- a/dusty/schemas/test_schema.py
+++ b/dusty/schemas/test_schema.py
@@ -1,0 +1,16 @@
+from schemer import Schema, Array
+
+test_suite_schema = Schema({
+    'name': {'type': basestring, 'required': True},
+    'command': {'type': basestring, 'required': True},
+    'default_args': {'type': basestring, 'default': ''}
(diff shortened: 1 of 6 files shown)


dusty/schemas/test_schema.py
```diff
@@ -0,0 +1,16 @@
...
+})
+
+test_schema = Schema({
+    'image': {'type': basestring, 'default': ''},
+    'build': {'type': basestring},
+    'services': {'type': Array(basestring), 'default': list},
+    'once': {'type': basestring},
+    'compose': {'type': dict, 'default': dict},
+    'suites': {'type': Array(test_suite_schema), 'default': list},
+    }, strict=False)
```
paetling: we probably want this schema to be strict.    it was only false so that I did not have to do this test_schema stuff as I was rushing through.  A non strict schema will allow you to have more keys than those specified above.  Which is probably not what we want.

thieman: +1

PR #205, opened 2015-06-14
Pull test base image if it doesn't already exist

thieman:
@paetling We need to do this ourselves whenever we're doing stuff directly through the Docker API. When we're using Compose, it has our backs.

### DIFF
--- a/dusty/systems/docker/testing_image.py
+++ b/dusty/systems/docker/testing_image.py
@@ -33,11 +33,21 @@ def _get_create_container_binds(split_volumes):
         binds_dict[volume_dict['host_location']] =  {'bind': volume_dict['container_location'], 'ro': False}
     return binds_dict
 
+def _ensure_image_exists(docker_client, image_name):
+    for image in docker_client.images():
+        if any(image_name in tag for tag in image['RepoTags']):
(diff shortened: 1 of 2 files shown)


dusty/systems/docker/testing_image.py
```diff
@@ -33,11 +33,21 @@ def _get_create_container_binds(split_volumes):
         binds_dict[volume_dict['host_location']] =  {'bind': volume_dict['container_location'], 'ro': False}
     return binds_dict
 
+def _ensure_image_exists(docker_client, image_name):
+    for image in docker_client.images():
+        if any(image_name in tag for tag in image['RepoTags']):
+            break
+    else:
```
paetling: when will python go into the else?

thieman: If the `for` doesn't encounter a `break`

thieman: > Loop statements may have an else clause; it is executed when the loop terminates through exhaustion of the list (with for) or when the condition becomes false (with while), but not when the loop is terminated by a break statement.

paetling:
:ghost: :cow:

PR #208, opened 2015-06-14
make tests handle libs as well as apps

paetling:
review @thieman


thieman:
Whole thing is desperately crying for a refactor. Code looks fine, though.

PR #209, opened 2015-06-15
use || since we have set -e

jsingle:
@thieman we gotta do this if we want to print this **essential** error message (because of `set -e`)


thieman:
lol k

PR #210, opened 2015-06-15
make sure that boot2docker is running before you run tests

paetling:
review @thieman   I think this should be the one other command other than run that ensures that the vm is running.  This is because it is a totally different flow then up and should be supported independently


jsingle:
Nice I was just about to submit the same PR :)

jsingle:
Hmmm do we want to just call `initialize_docker_vm()`?

paetling:
instead of doing what?

jsingle:
Lol nevermind I'm dumb.  deleted my other comment

PR #211, opened 2015-06-15
Consolidate the binaries, dusty -d runs daemon now

thieman:
@jsingle


jsingle:
Did you just grep for `dustyd` to find where we're referencing it?  This is all I can think of, obviously jenkins needs fixin too

PR #213, opened 2015-06-15
Js set home on demote

jsingle:
@thieman I was having trouble with demoted boot2docker running from the plist, since the demoted process still has HOME set to `/var/root`

### DIFF
--- a/dusty/subprocess.py
+++ b/dusty/subprocess.py
@@ -23,7 +23,7 @@ def _demote_to_user(user_name):
     return _demote
 
 def _set_demoted_home_dir(user_name):
-    home_dir = '/var/root' if user_name == 'root' else '/Users/{}'.format(user_name)
+    home_dir = os.path.expanduser('~{}'.format(user_name))
     os.environ['HOME'] = home_dir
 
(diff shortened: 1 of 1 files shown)


dusty/subprocess.py
```diff
@@ -17,10 +17,15 @@
 def _demote_to_user(user_name):
     def _demote():
         pw_record = pwd.getpwnam(user_name)
+        _set_demoted_home_dir(user_name)
         os.setgid(pw_record.pw_gid)
         os.setuid(pw_record.pw_uid)
     return _demote
 
+def _set_demoted_home_dir(user_name):
+    home_dir = '/var/root' if user_name == 'root' else '/Users/{}'.format(user_name)
```
thieman: This is the wrong way to do this. Use `os.path.expanduser`

thieman:
This doesn't feel like the right way to do this. Are there alternatives?

dusty/subprocess.py
jsingle: Ah that's nifty

thieman:
Well actually I guess this makes sense, just change it to use expanduser

jsingle:
There might be a better way that we should be demoting - where we actually login as a user, then run the shell args

PR #214, opened 2015-06-15
add in command that will ensure you are in the mounted directory before running tests

paetling:
review @jsingle

### DIFF
--- a/tests/unit/utils.py
+++ b/tests/unit/utils.py
@@ -8,9 +8,13 @@ def get_app_dusty_schema(doc):
         doc['image'] = ''
     if 'repo' not in doc:
         doc['repo'] = ''
+    if 'mount' not in doc:
+        doc['mount'] = ''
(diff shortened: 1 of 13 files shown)


jsingle:
Hmm I'm surprised this worked at all before

paetling:
reason it worked is because gcweb specifies an entrypoint of /gc/gcweb.  not all images do that.  So we should make sure we move you over

tests/unit/utils.py
```diff
@@ -8,9 +8,13 @@ def get_app_dusty_schema(doc):
         doc['image'] = ''
     if 'repo' not in doc:
         doc['repo'] = ''
+    if 'mount' not in doc:
+        doc['mount'] = ''
     return DustySchema(app_schema, doc)
```
thieman: I feel like this should die and our fixtures should be forced to be explicit about the shit they need. This seems overly magical.

PR #215, opened 2015-06-15
Pull specs repo after initial setup

thieman:
@jsingle Also fixes a bug where config warnings didn't get refreshed after running setup

### DIFF
--- a/dusty/commands/setup.py
+++ b/dusty/commands/setup.py
@@ -3,9 +3,10 @@ import subprocess
 import textwrap
 
 from ..payload import Payload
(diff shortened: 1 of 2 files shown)


dusty/commands/setup.py
```diff
@@ -61,4 +62,6 @@ def save_dusty_config(config):
     for key, value in config.iteritems():
```
jsingle: Should this function be named something different?  At this point I'm not sure `save_dusty_config` is a great description for what it does

thieman: +1

PR #218, opened 2015-06-15
allow for requests connecting to logs timeout

paetling:
review @thieman   feel free to suggest a better way

### DIFF
--- a/dusty/systems/docker/__init__.py
+++ b/dusty/systems/docker/__init__.py
@@ -37,7 +37,8 @@ def get_docker_client():
     env = get_docker_env()
     host, cert_path, tls_verify = env['DOCKER_HOST'], env['DOCKER_CERT_PATH'], env['DOCKER_TLS_VERIFY']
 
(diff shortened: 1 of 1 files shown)


dusty/commands/test.py
```diff
@@ -103,6 +104,18 @@ def _app_or_lib_compose_up(testing_spec, app_or_lib_name, app_or_lib_volumes, te
...
 
+# client.logs times out after 60 seconds.  The timeout is not configurable, this is a workaround
+def _get_client_logs(client, test_container_name, stdout=True, stderr=True, stream=True):
+    while True:
+        try:
+            val = client.logs(test_container_name, stdout=stdout, stderr=stderr, stream=stream)
+        except requests.exceptions.ReadTimeout as e:
+            continue
+        except:
+            raise
```
jsingle: why do we need this except/raise?

paetling: we don't

PR #221, opened 2015-06-15
Ae remove imgaes

paetling:
review @thieman

### DIFF
--- a/dusty/systems/docker/testing_image.py
+++ b/dusty/systems/docker/testing_image.py
@@ -49,7 +49,10 @@ def _make_installed_requirements_image(docker_client, base_image_tag, command, i
     create_container_binds = _get_create_container_binds(split_volumes)
 
     _ensure_image_exists(docker_client, base_image_tag)
-    docker_client.remove_image(image=image_name)
+    try:
+        docker_client.remove_image(image=image_name)
+    except:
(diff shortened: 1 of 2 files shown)


dusty/systems/docker/testing_image.py
```diff
@@ -5,6 +5,12 @@
 from ...compiler.compose import lib_install_commands_for_app_or_lib, container_code_path, get_volume_mounts
 from ...log import log_to_client
 
+def _remove_image_if_exists(docker_client, image_name):
+    try:
+        docker_client.remove_image(image=image_name)
+    except:
+        log_to_client('was not able to remove image {}'.format(image_name))
```
thieman: Seems like this function should really just be done inline

paetling: had it inline initially. but needed to do it two places.   if i do it with a function, i am not littering try excpets multiple places

thieman: yeah but exception handling isn't really a good candidate for throwing into a function imo. you're really just obfuscating what you're actually trying to do (`docker_client.remove_image` which is pretty straightforward) and adding some sneaky exception handling that isn't really hinted at by the name of the function

The "meat" of the function is only one line, and you're actually making it more complex and harder to reason about by abstracting it, so it seems like a bad fit

PR #223, opened 2015-06-16
Js specs class

jsingle:
@paetling do these changes seem reasonable?   There's probably a lot more existing code that could be refactored, but new code will be a little easier to write...

Also lots of tests need fixing :/

### DIFF
--- a/dusty/commands/run.py
+++ b/dusty/commands/run.py
@@ -71,8 +71,8 @@ def restart_apps_or_services(app_or_service_names=None, sync=True):
     if sync:
         if app_or_service_names:
             specs = spec_assembler.get_assembled_specs()
-            app_names = [app_name for app_name in app_or_service_names if app_name in specs['apps']]
-            rsync.sync_repos_by_app_name(specs, app_names)
+            specs_list = [specs['apps'][app_name] for app_name in app_or_service_names if app_name in specs['apps']]
+            rsync.sync_repos_by_specs(specs_list)
         else:
             rsync.sync_repos(spec_assembler.get_all_repos(active_only=True, include_specs_repo=False))
 
--- a/dusty/commands/run.py
+++ b/dusty/commands/run.py
@@ -69,9 +69,12 @@ def restart_apps_or_services(app_or_service_names=None, sync=True):
         log_to_client("Restarting all active containers associated with Dusty")
 
     if sync:
+        print "IN SYNC", app_or_service_names
         if app_or_service_names:
             specs = spec_assembler.get_assembled_specs()
+            print specs
             specs_list = [specs['apps'][app_name] for app_name in app_or_service_names if app_name in specs['apps']]
+            print specs_list
             rsync.sync_repos_by_specs(specs_list)
         else:
             rsync.sync_repos(spec_assembler.get_all_repos(active_only=True, include_specs_repo=False))
--- a/dusty/commands/run.py
+++ b/dusty/commands/run.py
@@ -69,12 +69,9 @@ def restart_apps_or_services(app_or_service_names=None, sync=True):
         log_to_client("Restarting all active containers associated with Dusty")
 
     if sync:
-        print "IN SYNC", app_or_service_names
         if app_or_service_names:
             specs = spec_assembler.get_assembled_specs()
-            print specs
             specs_list = [specs['apps'][app_name] for app_name in app_or_service_names if app_name in specs['apps']]
-            print specs_list
             rsync.sync_repos_by_specs(specs_list)
         else:
             rsync.sync_repos(spec_assembler.get_all_repos(active_only=True, include_specs_repo=False))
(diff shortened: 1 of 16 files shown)


dusty/schemas/base_schema_class.py
```diff
@@ -19,8 +22,70 @@ def __delitem__(self, key):
...
+    def get_specs_from_path(self, specs_path):
+        specs = {}
+        for key in ['bundles', 'apps', 'libs', 'services']:
+            specs[key] = {}
+            schema = _get_respective_schema(key)
+            key_path = os.path.join(specs_path, key)
+            for spec_path in glob.glob('{}/*.yml'.format(key_path)):
+                spec_name = os.path.splitext(os.path.split(spec_path)[-1])[0]
+                with open(spec_path, 'r') as f:
+                    spec = yaml.load(f.read())
```
paetling: names should be unique right?   Should we be enforcing that here?

paetling:
look in command/test.py and docker/testing_image.py for more files that can use this new system

dusty/schemas/base_schema_class.py
jsingle: This is basically old code copied from before.  I don't think it's an issue because the names are based on filenames, which have to be unique

paetling:
generally looks pretty good.  nice that we can get ride of these type checks throughout the code.  Are we only doing type checking for the lib, app differences?

dusty/schemas/base_schema_class.py
paetling: unique per directory.  But i thought our namespace was globally unique?

jsingle:
There are other cases where we pass type around as a function argument.  And there are other situations where it's nice to know the name as well, along with all its dict contents, in the same object

paetling:
Yeah.  I thought there were more places this would be needed also.  I am more asking if you have done some auditing to see if you have found most of the cases we care about

jsingle:
I haven't done that much auditing yet.  Wanted to check in with someone to make sure I wasn't doing something obviously dumb first, since combing for this stuff might be tedious

dusty/schemas/base_schema_class.py
jsingle: As in a lib can't have the same name as an app?  I thought we just needed uniqueness within a given type, and each spec of a given type lives in the same dir

paetling:
spec_assembler.py line 141.

paetling:
looks pretty good.  ping me again when it is all done for one more look over

dusty/commands/run.py
```diff
@@ -69,10 +69,13 @@ def restart_apps_or_services(app_or_service_names=None, sync=True):
...
 
     if sync:
+        print "IN SYNC", app_or_service_names
         if app_or_service_names:
             specs = spec_assembler.get_assembled_specs()
-            app_names = [app_name for app_name in app_or_service_names if app_name in specs['apps']]
-            rsync.sync_repos_by_app_name(specs, app_names)
+            print specs
+            specs_list = [specs['apps'][app_name] for app_name in app_or_service_names if app_name in specs['apps']]
+            print specs_list
```
paetling: remove the 2 prints

dusty/commands/test.py
```diff
@@ -104,7 +93,8 @@ def _app_or_lib_compose_up(testing_spec, app_or_lib_name, app_or_lib_volumes, te
     return '{}_{}_1'.format(_compose_project_name(app_or_lib_name), app_or_lib_name)
 
 def _run_tests_with_image(client, expanded_specs, app_or_lib_name, test_command):
-    testing_spec = _spec_for_service(app_or_lib_name, expanded_specs)['test']
+
+    testing_spec = expanded_specs.get_app_or_lib(app_or_lib_name)['test']
```
paetling: space

dusty/compiler/spec_assembler.py
```diff
@@ -138,46 +135,17 @@ def get_repo_of_app_or_library(app_or_library_name):
...
     if include_specs_repo:
         repos.add(get_specs_repo())
     specs = get_assembled_specs() if active_only else get_specs()
-    for type_key in ['apps', 'libs']:
-        for spec in specs[type_key].itervalues():
-            if 'repo' in spec:
-                repos.add(Repo(spec['repo']))
+    for spec in specs.get_apps_and_libs():
+        if 'repo' in spec:
+            repos.add(Repo(spec['repo']))
```
paetling: this repo check should no longer be required.  'repo' is a required field in schema

dusty/schemas/base_schema_class.py
```diff
@@ -19,8 +22,73 @@ def __delitem__(self, key):
...
+            return self._document['apps'][app_or_lib_name]
+        elif app_or_lib_name in self._document['libs']:
+            return self._document['libs'][app_or_lib_name]
+        raise KeyError('did not find app or service with name {}'.format(app_or_lib_name))
+
+    def get_apps_and_libs(self):
+        return [app for app in self._document['apps'].values()] + [lib for lib in self._document['libs'].values()]
+
+    def get_apps_and_services(self):
+        return [app for app in self._document['apps'].values()] + [serv for serv in self._document['services'].values()]
```
paetling: so you have any tests for this class. If not, you need some. (does not seem like you have any)

paetling:
This looks pretty good, some small stuff and requests for tests

dusty/schemas/base_schema_class.py
jsingle: ugh tests

PR #224, opened 2015-06-16
Fix error in scripts CLI help

thieman:
self-merge


nrschultz:
lol

thieman:
how does find and replace even work

thieman:
we just don't know

PR #225, opened 2015-06-16
Ae better nginx conf

paetling:
review @thieman I decided on values in list from this page: http://nginx.org/en/docs/beginners_guide.html

### DIFF
--- a/dusty/commands/setup.py
+++ b/dusty/commands/setup.py
@@ -1,6 +1,7 @@
 import pwd
 import subprocess
 import textwrap
+from os.path import isfile
 
 from ..payload import Payload
 from ..config import save_config_value, get_config_value, verify_mac_username, refresh_config_warnings
@@ -27,11 +28,20 @@ def _get_default_specs_repo():
     return _get_raw_input('Input the full name of your specs repo, e.g. github.com/gamechanger/example-dusty-specs: ')
 
 def _get_nginx_includes_dir():
-    _pretty_print_key_info(constants.CONFIG_NGINX_DIR_KEY)
-    default_nginx_config_value = get_config_value(constants.CONFIG_NGINX_DIR_KEY)
-    if _get_raw_input('Does your nginx config look for extra configs in the default location of {}? (y/n): '.format(default_nginx_config_value)).upper() == 'Y':
-        return default_nginx_config_value
-    return _get_raw_input('Input the path where your nginx config pulls extra configs: ')
+    nginx_include_dir = None
+    for nginx_conf_location in constants.NGINX_CONFIG_FILE_LOCATIONS:
+        file_location = '{}/nginx.conf'.format(nginx_conf_location)
+        if isfile(file_location):
+            with open(file_location) as f:
+                contents = f.read()
+                if 'include servers/*;' in contents:
+                    nginx_include_dir = '{}/servers'
+                    break
+
+    if nginx_includes_dir is None:
+        return _get_raw_input('You have non standard nginx config setup. Could not find an nginx.conf file that includes a directory. Please input the full path of the directory your nginx config includes. ')
+    else:
+        return nginx_include_dir
 
 def setup_dusty_config(mac_username=None, specs_repo=None, nginx_includes_dir=None):
     print "We just need to verify a few settings before we get started.\n"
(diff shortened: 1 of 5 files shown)


dusty/commands/setup.py
```diff
@@ -26,12 +27,24 @@ def _get_default_specs_repo():
...
+    for nginx_conf_location in constants.NGINX_CONFIG_FILE_LOCATIONS:
+        file_location = '{}/nginx.conf'.format(nginx_conf_location)
+        if isfile(file_location):
+            contents = _get_contents_of_file(file_location)
+            if 'include servers/*;' in contents:
+                nginx_includes_dir = '{}/servers'.format(nginx_conf_location)
+                break
+
+    if nginx_includes_dir is None:
+        return _get_raw_input('You have non standard nginx config setup. Could not find an nginx.conf file that includes a directory. Please input the full path of the directory your nginx config includes. ')
```
thieman: Textwrap this please

thieman: Also this block could probably be in the `else` of the `for`, right?

dusty/commands/setup.py
```diff
@@ -26,12 +28,35 @@ def _get_default_specs_repo():
...
+        return f.readlines()
+
+def _append_to_file(file_location, new_line):
+    with open(file_location, 'a') as f:
+        f.write(new_line)
+
+def _setup_nginx_config(nginx_conf_location):
+    include_path = '{}/servers'.format(nginx_conf_location)
+    if not isdir(include_path):
+        mkdir(include_path)
```
thieman: This is very likely to fail since setup runs as a nonprivileged user.

thieman: Actually that kind of ruins the whole update-it-for-them plan doesn't it

paetling: So the update it for them works because setup is done through dusty d which is run as sudo.

thieman: Sorry I forgot to turn off my internal dumbass setting today

paetling: :)

dusty/commands/setup.py
```diff
@@ -26,12 +28,35 @@ def _get_default_specs_repo():
...
+    _append_to_file('{}/nginx.conf'.format(nginx_conf_location), 'include servers/*;')
+    return include_path
+
 def _get_nginx_includes_dir():
```
thieman: Probably needs a less innocuous name now

dusty/commands/setup.py
```diff
@@ -26,12 +28,35 @@ def _get_default_specs_repo():
...
+
+def _append_to_file(file_location, new_line):
+    with open(file_location, 'a') as f:
+        f.write(new_line)
+
+def _setup_nginx_config(nginx_conf_location):
+    include_path = '{}/servers'.format(nginx_conf_location)
+    if not isdir(include_path):
+        mkdir(include_path)
+    _append_to_file('{}/nginx.conf'.format(nginx_conf_location), 'include servers/*;')
```
thieman: May want to surround this in newlines to be safe in case they don't already have a newline at the end of the file

dusty/commands/setup.py
```diff
@@ -26,12 +28,35 @@ def _get_default_specs_repo():
...
+            contents = _get_contents_of_file(file_location)
+            for line in contents:
+                if line.startswith('include'):
+                    include_folder = line.replace('include ', '').replace('/*;', '').replace('\n', '').strip()
+                    return '{}/{}'.format(nginx_conf_location, include_folder)
+            if _get_raw_input('\n'.join(textwrap.wrap('You have non standard nginx config setup. Can we add "include servers/*;" to the end of your nginx config file ({})? If you select no, your nginx forwarding will not work. (y/n) '.format(file_location), 80))).upper() == 'Y':
+                return _setup_nginx_config(nginx_conf_location)
+            else:
+                return ''
+    return _get_raw_input('\n'.join(textwrap.wrap('You have non standard nginx config setup. Could not find an nginx.conf file. Please input the full path of the directory your nginx config includes. ', 80)))
```
thieman: This is really jank if anyone makes it this far, wonder if there's a better way we could handle this

thieman:
Looks good on design

dusty/commands/setup.py
paetling: thats a good idea

PR #232, opened 2015-06-17
add in check when starting up dusty daemon to ensure tyou can talk to github over ssh

paetling:
review @thieman can't do it in preflight, because we are not ensured that username has been set in preflight


thieman:
:bear:

PR #235, opened 2015-06-17
add in message about dusty_logs to dusty up

paetling:
review @jsingle   got a request for this on monday. I think it could be a good idea (why is gcweb showing a 403)


jsingle:
:+1:

PR #237, opened 2015-06-17
Ae dusty status

paetling:
review @jsingle

### DIFF
--- a/dusty/commands/status.py
+++ b/dusty/commands/status.py
@@ -9,21 +9,13 @@ def _has_active_container(spec_type, service_name):
         return False
     return get_dusty_containers([service_name]) != []
 
-def _get_singular_spec_type(plural_spec_type):
-    if plural_spec_type == 'libs':
-        return 'lib'
-    elif plural_spec_type == 'apps':
-        return 'app'
-    elif plural_spec_type == 'services':
-        return 'service'
-
 def get_dusty_status():
     assembled_specs = get_assembled_specs()
     table = PrettyTable(["Name", "Type", "Has Active Container"])
     import logging
     logging.error(assembled_specs._document)
     for spec in assembled_specs.get_apps_libs_and_services():
-        spec_type = _get_singular_spec_type(spec.spec_type)
+        spec_type = spec.spec_type.rstrip('s')
         service_name = spec.name
         has_activate_container = _has_active_container(spec_type, service_name)
         table.add_row([service_name, spec_type, 'X' if has_activate_container else ''])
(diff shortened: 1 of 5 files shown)


paetling:
image

paetling:
image

dusty/commands/status.py
```diff
@@ -0,0 +1,19 @@
+from prettytable import PrettyTable
+
+from ..compiler.spec_assembler import get_assembled_specs
+from ..log import log_to_client
+from ..systems.docker import get_dusty_containers
+
+def _has_activate_container(spec_type, service_name):
```
jsingle: Shouldn't this be `_has_active_container`?

dusty/commands/status.py
```diff
@@ -0,0 +1,19 @@
...
+    if spec_type == 'libs':
+        return False
+    return get_dusty_containers([service_name]) != []
+
+def get_dusty_status():
+    assembled_specs = get_assembled_specs()
+    table = PrettyTable(["Name", "Type", "Has Active Container"])
+    for spec_type in ['apps', 'libs', 'services']:
+        for service_name, spec in assembled_specs[spec_type].iteritems():
```
jsingle: If you `itervalues`, spec.name will just give you its name, and spec.spec_type gives you the type

dusty/commands/status.py
paetling: yeah
good catch

jsingle:
How hard would it be to print the container ID too?

dusty/commands/status.py
```diff
@@ -0,0 +1,19 @@
...
+
+def _has_activate_container(spec_type, service_name):
+    if spec_type == 'libs':
+        return False
+    return get_dusty_containers([service_name]) != []
+
+def get_dusty_status():
+    assembled_specs = get_assembled_specs()
+    table = PrettyTable(["Name", "Type", "Has Active Container"])
+    for spec_type in ['app', 'lib', 'service']:
```
jsingle: Instead of doing this, I'd advocate writing a method that goes on `assembled_specs` class, which returns a list or a set of the specs you need.  Those will have `.name` and `.spec_type` attribute, which you will put in your table

paetling: cool.  will do that.

paetling:
What is the value of printing container ids, those are purely docker things?  I do not think it really helps them use dusty better.  Thoughts?

jsingle:
Well it's just useful for them to run various docker commands easily.  I definitely wouldn't say it's necessary - I suppose if people know how to use docker they could easily just `docker ps` anyway

paetling:
Yeah that is sort of my thought. If you want to use docker commands on containers, you should get the docker container by using docker ps.   If you are doing dusty commands you never need to explictily know the container id

PR #241, opened 2015-06-17
Ae fix up

paetling:
review @jsingle

### DIFF
--- a/dusty/commands/repos.py
+++ b/dusty/commands/repos.py
@@ -46,8 +46,9 @@ def update_managed_repos():
     log_to_client('Pulling latest updates for all active managed repos:')
     overrides = set(get_config_value(constants.CONFIG_REPO_OVERRIDES_KEY))
     specs_repo = get_specs_repo()
-    log_to_client('Updating managed copy of specs-repo before loading specs')
-    specs_repo.update_local_repo()
+    if not specs_repo.is_overridden:
+        log_to_client('Updating managed copy of specs-repo before loading specs')
+        specs_repo.update_local_repo()
     for repo in get_all_repos(active_only=True, include_specs_repo=False):
         if not repo.is_overridden:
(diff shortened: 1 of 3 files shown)


tests/unit/commands/repos_test.py
```diff
@@ -77,11 +77,11 @@ def test_override_repos_from_directory(self):
...
-        fake_update_local_repo.assert_called_once_with()
+        fake_update_local_repo.assert_has_calls([call(), call()])
 
     @patch('dusty.source.Repo.update_local_repo')
     def test_update_managed_repos_for_both(self, fake_update_local_repo):
         activate_bundle(['bundle-a'])
         activate_bundle(['bundle-b'])
         update_managed_repos()
-        fake_update_local_repo.assert_has_calls([call(), call()])
+        fake_update_local_repo.assert_has_calls([call(), call(), call()])
```
paetling: really weird that these did not start failing till i actually changed the values in here (these tests seem to never have been correct)

dusty/commands/repos.py
```diff
@@ -45,7 +45,10 @@ def update_managed_repos():
     copy to bring it up to date with the latest master."""
     log_to_client('Pulling latest updates for all active managed repos:')
     overrides = set(get_config_value(constants.CONFIG_REPO_OVERRIDES_KEY))
-    for repo in get_all_repos(active_only=True):
+    specs_repo = get_specs_repo()
+    log_to_client('Updating managed copy of specs-repo before loading specs')
+    specs_repo.update_local_repo()
```
jsingle: we only want to do this if it's managed, right?

tests/unit/commands/repos_test.py
jsingle: :space_invader:

dusty/commands/repos.py
paetling: that is a good point.

PR #245, opened 2015-06-18
Exception for diverging versions

zacharym:
@thieman is this along the lines of what you were thinking? It seems funny to only raise for this when if user runs with -v but it also seems funny to check before every command.


dusty/cli/__init__.py
```diff
@@ -96,6 +98,12 @@ def main():
     if '-d' in sys.argv:
         return run_daemon()
 
+    if '-v' in sys.argv:
```
thieman: This can be a normal client option I think, the `-d` handling is special because we need to bypass the client docopt entirely and use the daemon docopt instead.

dusty/cli/__init__.py
```diff
@@ -96,6 +98,12 @@ def main():
...
+        sock = _connect_to_daemon()
+        daemon_check = Payload(version.check_version, constants.VERSION)
+        daemon_version = _run_command(sock, daemon_check.serialize())
+        return
```
thieman: Also you're duplicating all this logic here, don't want to do that unless necessary

thieman:
Let's talk tomorrow, this is tricky

dusty/cli/__init__.py
zacharym: but what if the user wants to run only `dusty -v`? Basically what I was trying to avoid is adding another command specifically for this

zacharym:
going to close this and reopen later for more targeted PRs

PR #246, opened 2015-06-18
function requires a dict not a class

jsingle:
@thieman


thieman:
ugly but ok

jsingle:
Yeah I know.  There might be a better way to get at this

PR #247, opened 2015-06-18
Ae use lists

paetling:
review @thieman switch to using lists over a single line combined using &&s part 1


dusty/commands/script.py
```diff
@@ -31,10 +31,11 @@ def execute_script(app_name, script_name, script_arguments=[]):
     if script_spec is None:
         raise KeyError('No script found named {} in specs for app {}'.format(script_name, app_name))
 
+    base_script_command = '; '.join(script_spec['command'])
```
thieman: How hard is it to make it accept a string or a list of strings? Now that we have users we want to avoid backwards incompatible changes if we can

PR #254, opened 2015-06-18
Ae update readme

paetling:
review @jsingle


jsingle:
You probably want to try running it, make sure stuff still works.  But I think it should

PR #255, opened 2015-06-19
Make repo and mount optional in app specs

thieman:
@paetling Ran into a case when creating example specs where I wanted to run a Python fileserver, which is a one-liner requiring no code. https://github.com/gamechanger/example-dusty-specs/blob/master/apps/fileserver.yml

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,7 @@
 # Changelog
 
-## 0.1.2
+## 0.1.2 (In Progress)
+  * `repo` and `mount` keys in app specs are now optional
 
 ## 0.1.1 (June 18, 2015)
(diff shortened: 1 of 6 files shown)


docs/changelog.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
-## 0.1.2
+## 0.1.2 (In Progress)
+  * `repo` and `mount` keys in app specs are now optional
```
paetling: preface this with a discription like the 0.1.1 release

paetling:
clean out the default settings in test/unit/utils.py   Should not be setting repo or mount for anybody now.  Other than that, :ghost: :cow:

PR #257, opened 2015-06-19
Js improve socket smarts

jsingle:
@thieman the client can't know if it's "queued" (I don't think), so I'm just printing a message indicating that the command might hang until the daemon's ready


jsingle:
No idea how this could be causing this integration test to fail

PR #259, opened 2015-06-19
Multiple payload support

jsingle:
@paetling haven't tested that this works, but the idea look OK?

### DIFF
--- a/dusty/cli/__init__.py
+++ b/dusty/cli/__init__.py
@@ -102,6 +102,21 @@ def _connect_to_daemon():
     sock.settimeout(None)
     return sock
 
+def _run_payload(payload):
+    if payload.run_on_daemon:
+        sock = _connect_to_daemon()
+        if sock is None:
+            return 1
+        print 'The Dusty daemon will run your command once it\'s finished running previous commands'
+        try:
+            errored = _run_command(sock, result.serialize())
+        except KeyboardInterrupt:
+            print "Dusty Client stopping on KeyboardInterrupt..."
+            print "-----Dusty Daemon is still processing your last command!-----"
+        return errored
+    else:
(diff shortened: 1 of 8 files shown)


paetling:
seems to me that you would want to also make the system use the new always return payload.  Are you just waiting to do that in a new pr?

jsingle:
I wasn't going to but maybe that would make sense.  I was thinking that for commands that run entirely on the client, they can just run inline without doing a payload

dusty/payload.py
```diff
@@ -11,6 +12,9 @@ def __eq__(self, other):
             return self.fn == other.fn and self.args == other.args and self.kwargs == other.kwargs
         return False
 
+    def run(self):
+        self.fn(self.args, self.kwargs)
```
thieman: Don't these need various *'s

paetling:
I like the idea of standardizing on one way of handling commands (always returning a payload) and then the client can handle the rest

dusty/payload.py
jsingle: yep good catch

paetling: He is right. *args, **kwargs

jsingle:
OK I'll stick that in this PR then

paetling:
:ghost: :cow:

dusty/cli/__init__.py
```diff
@@ -102,6 +102,20 @@ def _connect_to_daemon():
     sock.settimeout(None)
     return sock
 
+def _run_payload(payload):
+    if payload.run_on_daemon:
+        sock = _connect_to_daemon()
+        if sock is None:
+            return 1
```
thieman: This is weird.

PR #260, opened 2015-06-19
support relative paths in build field

jsingle:
@paetling we claim that we do this in the gcweb spec.  Now we actually can


paetling:
:ghost: :cow:

PR #262, opened 2015-06-19
use plain property, not kwargs mess

jsingle:
@thieman this better?


thieman:
I think it's better than the alternative.

PR #271, opened 2015-06-22
installation docs

jsingle:
@thieman  I copied the command which curls v0.1.1 of the install script here.  Seems like we'd want our one-liner in this doc, but it's another thing we'd have to update whenever we bump the version

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -1,3 +1,63 @@
-# Installation
+# Installing Dusty
 
-Coming Soon
+## Requirements
+You must have the following installed in order to run Dusty:
+
+ * nginx
+ * Virtualbox
+ * boot2docker
+ * Docker Compose
+
+Not all versions of these programs will work with Dusty.  Dusty
+will warn you if you have a version installed that may be too old.
+We recommend brew-installing the latest versions of each of these
+programs.
+
+## Installation
+
+We provide a install script with each release.  This install script
+takes the following actions:
+
+* Downloads the Dusty binary, and places it at `/usr/local/bin/dusty`
+* Unloads any existing Dusty Daemon plist from `/System/Library/LaunchDaemons/org.gamechanger.dusty.plist`
+* Downloads our plist file to the same location
+* Runs dusty with `--preflight-only`, which will check some components of your system setup
+* Loads the plist that was downloaded, if the previous step was successful
+
+To download and run this install script, run
+`bash -c "`curl -L https://github.com/gamechanger/dusty/releases/download/0.1.1/install.sh`"`
+
+## Setup
+
+After you install Dusty, you should run `dusty setup` to do some
+necessary configuration.
+
+#### Username
+
+Dusty runs some commands as this user that you specify.  Usually the default
+here should be fine. Setting this to `root` is probably a **bad** idea.
+
+#### Specs Repo
+
+Dusty needs to know where your Dusty specs live.  You can enter a git
+repository here if you want to manage them that way (recommended).  You
+can also just specify a local file path.
+
+#### nginx Configuration
+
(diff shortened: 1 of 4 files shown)


docs/installation.md
```diff
@@ -1,3 +1,65 @@
...
+
+ * nginx
+ * Virtualbox
+ * boot2docker
+ * Docker Compose
+
+Not all versions of these programs will work with Dusty.  Dusty
+will warn you if you have a version installed that may be too old.
+We recommend brew-installing the latest versions of each of these
+programs.
```
thieman: As a kindness, we should list out the brew commands to achieve this.

docs/installation.md
```diff
@@ -1,3 +1,65 @@
...
+* Downloads our plist file to the same location
+* Runs dusty with `--preflight-only`, which will check some components of your system setup
+* Loads the plist that was downloaded, if the previous step was successful
+
+To download and run this install script, run
+```
+bash -c "`curl -L https://github.com/gamechanger/dusty/releases/download/0.1.1/install.sh`"
+```
+
+## Setup
```
thieman: This should be a subarticle.

docs/installation.md
```diff
@@ -1,3 +1,65 @@
...
+
+* Downloads the Dusty binary, and places it at `/usr/local/bin/dusty`
+* Unloads any existing Dusty Daemon plist from `/System/Library/LaunchDaemons/org.gamechanger.dusty.plist`
+* Downloads our plist file to the same location
+* Runs dusty with `--preflight-only`, which will check some components of your system setup
+* Loads the plist that was downloaded, if the previous step was successful
+
+To download and run this install script, run
+```
+bash -c "`curl -L https://github.com/gamechanger/dusty/releases/download/0.1.1/install.sh`"
```
thieman: Should mention that this is also how you upgrade.

thieman: Also if we're gonna have this here, add a note to the `Contributing` doc about how to cut new versions that mentions you need to update this.

docs/installation.md
```diff
@@ -1,3 +1,65 @@
-# Installation
+# Installing Dusty
 
-Coming Soon
+## Requirements
+You must have the following installed in order to run Dusty:
+
+ * nginx
+ * Virtualbox
```
thieman: I think boot2docker gives you this?

docs/installation.md
```diff
@@ -1,3 +1,65 @@
...
+The specific steps Dusty takes are:
+
+* Search standard nginx config file locations ('/usr/local/nginx/conf', '/etc/nginx', '/usr/local/etc/nginx') for `nginx.conf`
+* Search the first `nginx.conf` found for the `include` directive - this specifies a folder of additional nginx configuration files, which is where Dusty will write its configuration
+* If you don't include any folders, Dusty will ask you to let it add `include servers/*;` to the end of your Dusty config
+* If no `nginx.conf` is found, Dusty will let you know
+If Dusty can't setup its nginx config automatically, you just need to run
+```dusty config set nginx_includes_dir <folder>```
+Where `folder` is the absolute path to a folder from which your nginx will include `.conf`
+files when run.
```
thieman: This is a bug ugly block of text, can we pare it down somehow?

docs/installation.md
```diff
@@ -1,3 +1,65 @@
...
+#### Username
+
+Dusty runs some commands as this user that you specify.  Usually the default
+here should be fine. Setting this to `root` is probably a **bad** idea.
+
+#### Specs Repo
+
+Dusty needs to know where your Dusty specs live.  You can enter a git
+repository here if you want to manage them that way (recommended).  You
+can also just specify a local file path.
```
thieman: Should mention the example specs repo and link to the getting started section in case they want to run through that.

PR #272, opened 2015-06-22
Enabled? -> Activated?

nrschultz:
@thieman the command is `dusty activate` so it makes sense that afterwards, the bundle would be activated instead of enabled.


thieman:
So happy you changed the docs too, literally just got done working on those :-P

nrschultz:
Thats where I noticed it

PR #275, opened 2015-06-22
Ae test pull

paetling:
review @jsingle

### DIFF
--- a/dusty/cli/__init__.py
+++ b/dusty/cli/__init__.py
@@ -146,3 +146,4 @@ def main():
             errored = _run_payload(payload)
             if errored:
                 sys.exit(1)
+        sys.exit(0)
--- a/dusty/compiler/spec_assembler.py
+++ b/dusty/compiler/spec_assembler.py
@@ -149,3 +149,16 @@ def get_all_repos(active_only=False, include_specs_repo=True):
         if spec['repo']:
             repos.add(Repo(spec['repo']))
     return repos
+
+def _get_dependent_repos_for_app_or_library(app_or_library_name):
+    specs = get_specs()
+    spec = specs.get_app_or_lib(app_or_library_name)
+    repos = set()
+    for dependent_name in spec['depends'].get('apps') + spec['depends']['libs']:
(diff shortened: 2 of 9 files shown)


dusty/cli/__init__.py
```diff
@@ -146,3 +146,4 @@ def main():
             errored = _run_payload(payload)
             if errored:
                 sys.exit(1)
+        sys.exit(0)
```
jsingle: O.o

dusty/compiler/spec_assembler.py
```diff
@@ -149,3 +149,16 @@ def get_all_repos(active_only=False, include_specs_repo=True):
...
+
+def _get_dependent_repos_for_app_or_library(app_or_library_name):
+    specs = get_expanded_libs_specs()
+    spec = specs.get_app_or_lib(app_or_library_name)
+    repos = set()
+    for dependent_name in spec['depends']['libs']:
+        repos.add(Repo(specs.get_app_or_lib(dependent_name)['repo']))
+    return repos
+
+def get_all_repos_for_app_or_library(app_or_library_name):
```
jsingle: I feel like these functions are kind of testing-specific.  It's really getting all repos that are necessary for the test of an app or lib.  If we were getting all the repos necessary for running an app, we'd also want downstream app dependencies, for example.  If they're test specific, they should probably have different names and maybe go in a different file.  What do you think?

paetling: I think this is poorly name but is not just for tests.  It is more of, get_all_repos_for_app_or_library_container. If app1 depends on app2, it does not really care what repo app2 depends on, it just needs to wait for app2 to spin up before it spins up.  That is why I put them here.   What if i add _container to the end of thes function names?

jsingle: True.  However since libraries don't have containers on their own, it's still slightly off, no?  They have their own containers for testing purposes (again testing-specific)

paetling: that is true. The reason i put these here is the two functions above.  Which do similar stuff. gonna put them in test.py

jsingle:
Looks good :palm_tree:

PR #278, opened 2015-06-23
Touch up the FAQ a bit

thieman:
@jsingle


jsingle:
Proofreading ftw :8ball:

PR #281, opened 2015-06-23
adds test for payload client/daemon version validation

zacharym:
@jsingle cc @thieman

### DIFF
--- a/dusty/payload.py
+++ b/dusty/payload.py
@@ -1,5 +1,6 @@
 import cPickle
-from constants import VERSION
+
+from .constants import VERSION
 
 class Payload(object):
     def __init__(self, fn, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


dusty/payload.py
```diff
@@ -13,6 +15,8 @@ def __eq__(self, other):
         return False
 
     def run(self):
+        if self.client_version != VERSION:
+            raise RuntimeError("Dusty daemon is running version: {}, and client is running version: {}".format(VERSION, self.client_version))
```
jsingle: this is not the method the daemon uses to actually run the Payload.  This is just a helper function, which is actually just used by the client ATM.  Check the daemon code to see how the daemon should check the client_version

dusty/payload.py
```diff
@@ -1,9 +1,11 @@
 import cPickle
+from constants import VERSION
```
thieman: Relative imports always

zacharym:
@jsingle @thieman this is updated

jsingle:
:rocket: assuming you've tested it

PR #285, opened 2015-06-23
Zm dusty  v

zacharym:
@paetling cc @thieman

### DIFF
--- a/dusty/cli/__init__.py
+++ b/dusty/cli/__init__.py
@@ -24,6 +24,7 @@ Commands:
 
 Options:
   -d    Run the Dusty daemon
+  -v    Print the installed dusty version
 
 For help on a specific command, provide the '-h' flag to the command, e.g. 'dusty repos -h'
 """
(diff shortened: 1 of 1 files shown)


dusty/cli/__init__.py
```diff
@@ -125,7 +126,10 @@ def main():
     args = docopt(__doc__, options_first=True)
     command, command_args = args['<command>'], args['<args>']
     if command is None:
-        print __doc__.strip()
+        if '-v' in sys.argv:
+            print 'Dusty version {}'.format(constants.VERSION)
```
paetling: why not just check this first like the -d command?

paetling: @thieman i do not remember, just want to make sure. -v is for the client, version is for the deamon (that would make sense)

zacharym: I was thinking that it might be possible to have -v in the argv's in some other context. Do you think I should move it to above?

paetling: maybe the best way is to check that -v is the only thing provided (no commands) and do it before anything else
?

dusty/cli/__init__.py
```diff
@@ -121,6 +122,10 @@ def main():
     # we can process daemon-specific options
     if '-d' in sys.argv:
         return run_daemon()
+    if len(sys.argv) == 2 and sys.argv[1] == '-v':
```
paetling: i would favor actually checking <commands> more explicit that way.

paetling: commands

paetling:
looks good

dusty/cli/__init__.py
thieman: This needs to be shuffled around a little.

PR #286, opened 2015-06-23
Js restart repos

jsingle:
@thieman


dusty/cli/restart.py
```diff
@@ -6,21 +6,25 @@
...
 
 Options:
-  --no-sync    If provided, Dusty will not sync repos used by
-               services being restarted prior to the restart.
-  <services>   If provided, Dusty will only restart the given
-               services. Otherwise, all currently running
-               services are restarted.
+  --no-sync       If provided, Dusty will not sync repos used by
+                  services being restarted prior to the restart.
+  --repos <repos> If provided, Dusty will restart any containers
```
thieman: Keep 4 spaces padding please

dusty/cli/restart.py
```diff
@@ -6,21 +6,25 @@
...
 from docopt import docopt
 
 from ..payload import Payload
-from ..commands.run import restart_apps_or_services
+from ..commands.run import restart_apps_or_services, restart_apps_by_repo
 
 def main(argv):
     args = docopt(__doc__, argv)
+    if args['--repos']:
+        return Payload(restart_apps_by_repo, args['<repos>'], sync=not args['--no-sync'])
```
thieman: Should abstract the common args into a dict and then include it in both payloads with **

dusty/commands/run.py
```diff
@@ -85,3 +86,13 @@ def restart_apps_or_services(app_or_service_names=None, sync=True):
             rsync.sync_repos(spec_assembler.get_all_repos(active_only=True, include_specs_repo=False))
 
     compose.restart_running_services(app_or_service_names)
+
+def restart_apps_by_repo(repo_names, sync=True):
+    all_active_repos = spec_assembler.get_all_repos()
```
thieman: This isn't only getting active repos

dusty/commands/run.py
```diff
@@ -85,3 +86,13 @@ def restart_apps_or_services(app_or_service_names=None, sync=True):
...
+    resolved_repos = set([Repo.resolve(all_active_repos, repo_name) for repo_name in repo_names])
+    specs = spec_assembler.get_assembled_specs()
+    apps_with_repos = set()
+    for app_spec in specs['apps'].itervalues():
+        if spec_assembler.get_same_container_repos_from_spec(app_spec).intersection(resolved_repos):
+            apps_with_repos.add(app_spec.name)
+    restart_apps_or_services(list(apps_with_repos), sync=sync)
```
thieman: Is this `list` cast necessary?

dusty/commands/run.py
```diff
@@ -85,3 +86,13 @@ def restart_apps_or_services(app_or_service_names=None, sync=True):
...
 
     compose.restart_running_services(app_or_service_names)
+
+def restart_apps_by_repo(repo_names, sync=True):
+    all_active_repos = spec_assembler.get_all_repos()
+    resolved_repos = set([Repo.resolve(all_active_repos, repo_name) for repo_name in repo_names])
+    specs = spec_assembler.get_assembled_specs()
+    apps_with_repos = set()
+    for app_spec in specs['apps'].itervalues():
+        if spec_assembler.get_same_container_repos_from_spec(app_spec).intersection(resolved_repos):
```
thieman: This is likely to break at some point because it explicitly depends on the return type being a set. I think most Python programmers are used to sequences being used interchangeably, it wouldn't surprise me if someone changed it to a list at some point.

This should be tested, can you write integration tests for it?

jsingle: So should I explicitly cast this as a set?  That seems silly to do since it's currently a set

dusty/commands/run.py
jsingle: Likely not.  Hadn't traced downstream stuff before, and it does seem like everything's interchangeable

jsingle:
Will add tests (probably unit + integration)

PR #287, opened 2015-06-23
add version command

zacharym:
@jsingle cc @thieman

### DIFF
--- a/dusty/commands/version.py
+++ b/dusty/commands/version.py
@@ -1,0 +1,5 @@
+from ..constants import VERSION
+from ..log import log_to_client
+
+def version():
+    log_to_client('Dusty version: {}'.format(VERSION))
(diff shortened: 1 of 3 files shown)


dusty/commands/version.py
```diff
@@ -0,0 +1,5 @@
+from ..constants import VERSION
+from ..log import log_to_client
+
+def version():
+    log_to_client('Dusty version: {}'.format(VERSION))
```
jsingle: Let's make it clear that it's the *daemon* version

jsingle:
I'm not the one with the vision for the separate `dusty -v` and `dusty version` commands.  Seems to me like we should print out the client version as well in this command

jsingle:
Also it would be kind of funny to run `dusty version` and get a RuntimeError saying the client and daemon versions are different lol.  Not sure if that's worth addressing

dusty/commands/version.py
jsingle: DAEEEMON

thieman: +1

PR #288, opened 2015-06-23
compile args via docopt

zacharym:
@thieman @paetling the saga continues

### DIFF
--- a/dusty/cli/__init__.py
+++ b/dusty/cli/__init__.py
@@ -124,12 +124,12 @@ def main():
         return run_daemon()
 
     args = docopt(__doc__, options_first=True)
+    if args['-v']:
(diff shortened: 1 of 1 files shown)


thieman:
:bear:

dusty/cli/__init__.py
```diff
@@ -124,12 +124,12 @@ def main():
         return run_daemon()
 
     args = docopt(__doc__, options_first=True)
+    if args['-v']:
+        print 'Dusty version {}'.format(constants.VERSION)
+        sys.exit(0)
```
paetling: this fixes it because of docopt?   like if a did dusty logs -v it would not show up in args -v?

zacharym: that is correct

PR #290, opened 2015-06-23
Js backup persist

jsingle:
@paetling

### DIFF
--- a/dusty/cli/disk.py
+++ b/dusty/cli/disk.py
@@ -18,7 +18,8 @@ Commands:
 from docopt import docopt
 
 from ..payload import Payload
-from ..commands.disk import inspect_vm_disk, cleanup_inactive_containers, cleanup_images
+from ..commands.disk import (inspect_vm_disk, cleanup_inactive_containers, cleanup_images,
+                             backup, restore)
 
 def main(argv):
     args = docopt(__doc__, argv)
@@ -28,3 +29,7 @@ def main(argv):
         return Payload(cleanup_inactive_containers)
     elif args['cleanup_images']:
         return Payload(cleanup_images)
+    elif args['backup']:
+        return Payload(backup)
+    elif args['restore']:
+        return Payload(restore)
--- a/dusty/cli/disk.py
+++ b/dusty/cli/disk.py
@@ -4,11 +4,15 @@ Usage:
   disk inspect
   disk cleanup_containers
   disk cleanup_images
+  disk backup
+  disk restore
 
 Commands:
   inspect             Prints VM disk usage information
   cleanup_containers  Cleans docker containers that have exited
   cleanup_images      Removes docker images that can be removed without the --force flag
+  backup              Backs up the /persist directory on your boot2docker to your local file system
+  restore             Restores a backed up /persist directory
 """
 
 from docopt import docopt
--- a/dusty/cli/disk.py
+++ b/dusty/cli/disk.py
@@ -4,8 +4,8 @@ Usage:
   disk inspect
   disk cleanup_containers
   disk cleanup_images
-  disk backup
-  disk restore
+  disk backup <destination-path>
+  disk restore <source-path>
 
 Commands:
   inspect             Prints VM disk usage information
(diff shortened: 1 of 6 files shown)


dusty/commands/disk.py
```diff
@@ -1,28 +1,57 @@
...
+
+def _backup_modified_time():
+    return time.ctime(os.path.getmtime(constants.LOCAL_BACKUP_DIR))
+
+def backup():
+    _ensure_backup_dir_exists()
+    log_to_client("Ensuring virtualbox VM is running")
+    ensure_docker_vm_is_started()
+    log_to_client("Syncing data from your VM...")
+    sync_local_path_from_vm(constants.LOCAL_BACKUP_DIR, constants.VM_PERSIST_DIR)
```
paetling: is it worth having a warning around here because this will blow away your persist directory.  Something like, hey this will blow shit away, are you sure you want to do this?

paetling: I meant this about restore

jsingle: Good idea

jsingle:
Going to add user-specified destination for `backup`

dusty/cli/disk.py
```diff
@@ -24,3 +30,13 @@ def main(argv):
...
+    elif args['backup']:
+        path = os.path.abspath(args['<destination>'])
+        return Payload(backup, path)
+    elif args['restore']:
+        path = os.path.abspath(args['<source>'])
+        print "Warning: this will overwrite the /persist directory on your VM with the contents of {}".format(path)
+        if raw_input("Continue? (y/n) ").strip().upper() == 'Y':
+            return Payload(restore, path)
+        else:
+            print "Restore cancelled"
```
paetling: log_to_user. Have to do raw_input, but should still use log_to_user here

dusty/commands/disk.py
```diff
@@ -1,28 +1,59 @@
...
 
+def _full_destination_dir(destination_path):
+    return os.path.join(destination_path, constants.LOCAL_BACKUP_DIR)
+
+def _ensure_backup_dir_exists(destination_path):
+    if not os.path.exists(destination_path):
+        os.makedirs(destination_path)
+
+def backup(path):
+    destination_path = _full_destination_dir(path)
```
paetling: why are we joining on local_backup_dir and not just saving to where they tell us to?

dusty/commands/disk.py
```diff
@@ -1,28 +1,59 @@
...
+    _ensure_backup_dir_exists(destination_path)
+    log_to_client("Ensuring virtualbox VM is running")
+    ensure_docker_vm_is_started()
+    set_read_persist_permission()
```
paetling: what does this command do?

dusty/commands/disk.py
```diff
@@ -1,28 +1,59 @@
...
+    log_to_client("Syncing data from your VM to {}...".format(destination_path))
+    sync_local_path_from_vm(destination_path, constants.VM_PERSIST_DIR)
+    set_mac_user_ownership(destination_path)
```
paetling: and this one

dusty/commands/disk.py
jsingle: I could go either way on this.  If we add `dusty-backup` to the folder path, then when a user does `dusty disk backup ~/`, for example, he'll get a folder called `~/dusty-backup` containing the stuff, instead of a bunch like `~/coreMongo`, `~/coreRedis`, etc

dusty/commands/disk.py
paetling: why do we need this?

dusty/commands/disk.py
paetling: ehh. I sort of prefer the method where we allow the user to hang themselves instead of adding an extra layer, but this does add some amount of protection.  Do what you feel

dusty/commands/disk.py
jsingle: mongo sets permissions that sometimes look like `-rw------- root root`, so you can't even read the files to copy them

dusty/commands/disk.py
jsingle: I feel like otherwise the process would be something like `mkdir dusty-backup; dusty disk backup .` so it makes sense to save the extra command, and also prevent them from doing something dumb

dusty/cli/disk.py
jsingle: Why is log_to_user better in this situation?

paetling: just the standard way we send messages to users throught the daemon and the client

dusty/commands/disk.py
paetling: cool

PR #298, opened 2015-06-24
Zm repos integration

zacharym:
@jsingle cc @thieman


tests/testcases.py
```diff
@@ -136,13 +136,18 @@ def run_command(self, args, fake_exit, fake_exec_docker):
         self._clear_stdout()
         return result
 
-    def _set_up_fake_local_repo(self):
-        repo = git.Repo.init('/tmp/fake-repo')
-        with open('/tmp/fake-repo/README.md', 'w') as f:
-            f.write('# Fake Repo')
-        repo.index.add(['/tmp/fake-repo/README.md'])
+    def set_up_fake_local_repo(self, path='/tmp/fake-repo', short_name='Fake Repo'):
+        if path is None:
```
jsingle: why do we need this when we're using a default arg?

tests/testcases.py
```diff
@@ -136,13 +136,18 @@ def run_command(self, args, fake_exit, fake_exec_docker):
...
+            path = '/tmp/fake-repo'
+        repo = git.Repo.init(path)
+        with open('{}/README.md'.format(path), 'w') as f:
```
jsingle: use `os.path.join`

tests/testcases.py
zacharym: I'm not always using the default args

thieman: This still makes no sense, right?

tests/integration/cli/repos_test.py
```diff
@@ -0,0 +1,69 @@
...
+        result = self.run_command('repos list')
+        self.assertInSameLine(result, 'fake-repo', '/tmp/fake-repo')
+        self.assertNotInSameLine(result, 'fake-repo', '/tmp/fake-repo-1')
+        self.run_command('repos override fake-repo /tmp/fake-repo-1')
+        result = self.run_command('repos list')
+        self.assertInSameLine(result, 'tmp/fake-repo', 'fake-repo', '/tmp/fake-repo-1')
+
+    def test_repos_manage(self):
+        result = self.run_command('repos list')
+        self.assertNotInSameLine(result, 'fake-repo', '/tmp/fake-repo-1')
```
thieman: You don't need to this stuff like this every time you use it. Have one test that'll catch if it's broken, then just use it freely in other places.

tests/testcases.py
```diff
@@ -136,13 +136,18 @@ def run_command(self, args, fake_exit, fake_exec_docker):
...
+            f.write('# {}'.format(short_name))
+        repo.index.add([os.path.join(path, 'README.md')])
         repo.index.commit('Initial commit')
 
+    def tear_down_fake_local_repo(self, path='/tmp/fake-repo'):
+        shutil.rmtree(path)
```
thieman: This is just obfuscating a simple stdlib function, no reason to have it.

tests/integration/cli/repos_test.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def test_repos_update(self):
+        self.run_command('bundles activate busyboxa')
+        git_repo = git.Repo('/tmp/fake-repo')
+        check_output(['touch', '/tmp/fake-repo/foo'])
+        git_repo.index.add(['/tmp/fake-repo/foo'])
+        git_repo.index.commit('Second commit')
+        self.assertFalse(path.isfile('{}/tmp/fake-repo/foo'.format(REPOS_DIR)))
+        self.run_command('repos update')
+        self.assertTrue(path.isfile('{}/tmp/fake-repo/foo'.format(REPOS_DIR)))
```
thieman: Need to replace all instances of this pattern with `os.path.join`

tests/integration/cli/repos_test.py
```diff
@@ -0,0 +1,69 @@
...
+        busybox_single_app_bundle_fixture(num_bundles=1)
+        self.set_up_fake_local_repo(path='/tmp/fake-repo-1', short_name='fake-repo-1')
+
+    def tearDown(self):
+        super(TestReposCLI, self).tearDown()
+        self.tear_down_fake_local_repo(path='/tmp/fake-repo-1')
+
+    def test_repos_list(self):
+        result = self.run_command('repos list')
+        self.assertInSameLine(result, 'Full Name', 'Short Name', 'Local Override')
```
thieman: Don't need this line

tests/integration/cli/repos_test.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertInSameLine(result, 'github.com/gamechanger/dusty-example-specs', 'dusty-example-specs', self.overridden_specs_path)
+        self.assertInSameLine(result, 'tmp/fake-repo', 'fake-repo')
```
thieman: This is just checking whether `tmp/fake-repo` is in a line, you may need a more sophisticated assert function

tests/integration/cli/repos_test.py
```diff
@@ -0,0 +1,69 @@
+from os import path
+import shutil
+from subprocess import check_output
+
+import git
+
+from dusty.constants import REPOS_DIR
```
thieman: Should probably override this to a temp dir like we do in a few other places. As long as we do module-level imports everywhere, should get picked up fine.

tests/integration/cli/repos_test.py
```diff
@@ -0,0 +1,69 @@
...
+from ...fixtures import busybox_single_app_bundle_fixture
+from ...testcases import DustyIntegrationTestCase
+
+
+class TestReposCLI(DustyIntegrationTestCase):
+    def setUp(self):
+        super(TestReposCLI, self).setUp()
+        busybox_single_app_bundle_fixture(num_bundles=1)
+        self.set_up_fake_local_repo(path='/tmp/fake-repo-1', short_name='fake-repo-1')
```
thieman: Should be using `tempfile.mkdtemp` instead of this, I don't think it's always valid to assume `/tmp` is a writable location.

tests/integration/cli/repos_test.py
```diff
@@ -0,0 +1,69 @@
...
+        self.set_up_fake_local_repo(path='/tmp/from/fake-repo', short_name='fake-repo')
+        result = self.run_command('repos list')
+        self.assertNotInSameLine(result, 'fake-repo', '/tmp/from/fake-repo-1')
+        self.assertNotInSameLine(result, 'tmp/from/fake-repo', 'fake-repo')
+        self.run_command('repos from /tmp/from')
+        result = self.run_command('repos list')
+        self.assertInSameLine(result, '/tmp/fake-repo', 'fake-repo', 'tmp/from/fake-repo')
+        self.run_command('repos manage /tmp/fake-repo')
+        result = self.run_command('repos list')
+        shutil.rmtree('/tmp/from')
```
thieman: It isn't sufficient to do cleanup in the test. You need to do it in a tearDown function to ensure (well, in most cases) that it gets run.

tests/testcases.py
```diff
@@ -136,13 +136,18 @@ def run_command(self, args, fake_exit, fake_exec_docker):
         self._clear_stdout()
         return result
 
-    def _set_up_fake_local_repo(self):
-        repo = git.Repo.init('/tmp/fake-repo')
-        with open('/tmp/fake-repo/README.md', 'w') as f:
-            f.write('# Fake Repo')
-        repo.index.add(['/tmp/fake-repo/README.md'])
+    def set_up_fake_local_repo(self, path='/tmp/fake-repo', short_name='Fake Repo'):
```
thieman: `short_name` arg seems pointless

tests/integration/cli/repos_test.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def test_repos_update(self):
+        self.run_command('bundles activate busyboxa')
+        git_repo = git.Repo('/tmp/fake-repo')
+        check_output(['touch', '/tmp/fake-repo/foo'])
```
thieman: `check_call` would make more sense here

tests/testcases.py
zacharym: lol, yes

tests/integration/cli/repos_test.py
zacharym: Is it not testing that list displays these things? I'm happy to kill it though.

tests/integration/cli/repos_test.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class TestReposCLI(DustyIntegrationTestCase):
+    def setUp(self):
+        super(TestReposCLI, self).setUp()
+        busybox_single_app_bundle_fixture(num_bundles=1)
+        self.run_command('bundles activate busyboxa')
+        self.temp_repos_dir = mkdtemp()
+        self.old_repos_dir = constants.REPOS_DIR
+        constants.REPOS_DIR = self.temp_repos_dir
```
jsingle: Usually you don't want to set values of module variables like this.  I feel like we have other code in tests that you could look at for accomplishing something similar

jsingle: Actually I just blamed travis doing the same thing, so maybe not haha

tests/integration/cli/repos_test.py
```diff
@@ -0,0 +1,72 @@
...
+        super(TestReposCLI, self).tearDown()
+
+    def test_repos_list(self):
+        result = self.run_command('repos list')
+        self.assertInSameLine(result, 'github.com/gamechanger/dusty-example-specs', 'dusty-example-specs', self.overridden_specs_path)
+        self.assertInSameLine(result, self.fake_local_repo_location, 'fake-repo')
+
+    def test_repos_override(self):
+        result = self.run_command('repos list')
+        self.assertInSameLine(result, 'fake-repo', self.fake_local_repo_location)
```
jsingle: I'm confused about the naming between `fake_local_repo_location` and `fake_local_repo_location_1`

tests/integration/cli/repos_test.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def test_repos_manage(self):
+        self.run_command('repos override fake-repo {}'.format(self.fake_local_repo_location_1))
+        result = self.run_command('repos list')
+        self.assertInSameLine(result, self.fake_local_repo_location, 'fake-repo', self.fake_local_repo_location)
+        result = self.run_command('repos manage fake-repo')
+        self.assertNotInSameLine(result, 'fake-repo', self.fake_local_repo_location_1)
+        self.assertInSameLine(result, 'fake-repo', self.fake_local_repo_location)
+
+    def test_repos_from(self):
```
jsingle: Does it make sense to also test an additional repo that isn't in the "from" directory, to make sure it's left alone?

jsingle:
Lookin pretty good to me

tests/integration/cli/repos_test.py
thieman: I don't think Mock gives you a good way to do this with *values*, though I may be wrong

tests/integration/cli/repos_test.py
```diff
@@ -0,0 +1,72 @@
...
+        self.assertInSameLine(result, self.fake_local_repo_location, 'fake-repo', self.fake_local_repo_location)
+        result = self.run_command('repos manage fake-repo')
+        self.assertNotInSameLine(result, 'fake-repo', self.fake_override_repo_location)
+        self.assertInSameLine(result, 'fake-repo', self.fake_local_repo_location)
+
+    def test_repos_from(self):
+        self.run_command('repos from {}'.format(self.fake_from_dir))
+        result = self.run_command('repos list')
+        self.assertInSameLine(result, self.fake_local_repo_location, 'fake-repo', self.fake_from_repo_location)
+        self.run_command('repos manage {}'.format(self.fake_local_repo_location))
```
thieman: This looks like cleanup. **You can't do cleanup in testing functions.**

PR #300, opened 2015-06-25
Js restart integration

jsingle:
@thieman @zacharym

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -57,7 +57,7 @@ def specs_fixture_with_depends():
                             'mount': '/app/a',
                             'scripts': [{'description': 'A script description',
                                         'command': ['ls /'],
-                                        'name': 'example'}]
+                                        'name': 'example'}],
                             'depends': {
                                 'libs': ['lib-a']
                             }})
@@ -66,7 +66,7 @@ def specs_fixture_with_depends():
                             'mount': '/app/b',
                             'scripts': [{'description': 'A script description',
                                         'command': ['ls /'],
-                                        'name': 'example'}]
+                                        'name': 'example'}],
                             'depends': {
                                 'apps': ['app-a'],
                                 'libs': ['lib-b']
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -53,6 +53,9 @@ def specs_fixture_with_depends():
     _write('bundle', 'bundle-a', {'description': 'Bundle A', 'apps': ['appa']})
     _write('bundle', 'bundle-b', {'description': 'Bundle B', 'apps': ['appb']})
     _write('app', 'appa', {'repo': '/tmp/app-a',
+                            'commands': {
+                                'always': ['sleep 1000']
+                            },
                             'image': 'busybox',
                             'mount': '/app/a',
                             'scripts': [{'description': 'A script description',
@@ -62,6 +65,9 @@ def specs_fixture_with_depends():
                                 'libs': ['lib-a']
                             }})
     _write('app', 'appb', {'repo': '/tmp/app-b',
+                            'commands': {
+                                'always': ['sleep 1000']
+                            },
                             'image': 'busybox',
                             'mount': '/app/b',
                             'scripts': [{'description': 'A script description',
@@ -71,9 +77,6 @@ def specs_fixture_with_depends():
                                 'apps': ['appa'],
                                 'libs': ['lib-b']
                             }})
-    _write('app', 'appc', {'repo': '/tmp/app-c',
-                            'image': 'busybox',
-                            'mount': '/app/c',})
     _write('lib', 'lib-a', {'repo': '/tmp/lib-a',
                             'mount': '/lib/a',})
     _write('lib', 'lib-b', {'repo': '/tmp/lib-b',
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -50,9 +50,9 @@ def basic_specs_fixture():
     _write('service', 'service-a', {'image': 'service/a'})
 
 def specs_fixture_with_depends():
-    _write('bundle', 'bundle-a', {'description': 'Bundle A', 'apps': ['app-a']})
-    _write('bundle', 'bundle-b', {'description': 'Bundle B', 'apps': ['app-b']})
-    _write('app', 'app-a', {'repo': '/tmp/app-a',
+    _write('bundle', 'bundle-a', {'description': 'Bundle A', 'apps': ['appa']})
+    _write('bundle', 'bundle-b', {'description': 'Bundle B', 'apps': ['appb']})
+    _write('app', 'appa', {'repo': '/tmp/app-a',
                             'image': 'app/a',
                             'mount': '/app/a',
                             'scripts': [{'description': 'A script description',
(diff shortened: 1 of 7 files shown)


tests/integration/cli/restart_test.py
```diff
@@ -0,0 +1,71 @@
...
+from dusty.source import Repo
+from ...testcases import DustyIntegrationTestCase
+from ...fixtures import specs_fixture_with_depends
+
+class TestRestartCLI(DustyIntegrationTestCase):
+    def setUp(self):
+        super(TestRestartCLI, self).setUp()
+        specs_fixture_with_depends()
+        for repo in get_all_repos(include_specs_repo=False):
+            self._set_up_fake_local_repo(path=repo.remote_path)
```
zacharym: might want to rename this if it's not private anymore.. i guess it is private to the class not sure what convention is on that

tests/integration/cli/restart_test.py
```diff
@@ -0,0 +1,71 @@
...
+        self.run_command('bundles activate bundle-a bundle-b')
+        self.run_command('up')
+        time.sleep(1)
```
zacharym: why is this required?

tests/integration/cli/restart_test.py
thieman: It's still private, it's only being used from within the class.

tests/integration/cli/restart_test.py
```diff
@@ -0,0 +1,71 @@
...
+            self.run_command('stop')
+        except Exception:
+            pass
+        super(TestRestartCLI, self).tearDown()
+
+    @nottest
+    def container_has_restarted(self, app_name):
+        inspected = self.inspect_container(app_name)
+        start_time = dateutil.parser.parse(inspected['State']['StartedAt']).replace(tzinfo=None)
+        return start_time > self.up_complete_time
```
thieman: This should be an assert function defined in the test case.

tests/integration/cli/restart_test.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def test_restart_one(self):
+        self.run_command('restart appa')
+        self.assertTrue(self.container_has_restarted('appa'))
+        self.assertTrue(not self.container_has_restarted('appb'))
```
thieman: Much more idiomatic to just use `assertFalse`

tests/integration/cli/restart_test.py
zacharym: oh for `has_restarted`.. hmm

thieman: Doesn't `StartedAt` give us much more than second precision?

tests/integration/cli/restart_test.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def test_restart_by_app_repo(self):
+        self.run_command('restart --repos app-a')
+        self.assertTrue(self.container_has_restarted('appa'))
+        self.assertTrue(not self.container_has_restarted('appb'))
+
+    def test_restart_by_lib_repo(self):
+        self.run_command('restart --repos lib-a')
+        self.assertTrue(self.container_has_restarted('appa'))
+        self.assertTrue(self.container_has_restarted('appb'))
```
zacharym: Might be worth it to add an app that doesn't depend on this lib and confirm that it hasn't restarted

thieman: +1

tests/fixtures.py
```diff
@@ -49,6 +49,44 @@ def basic_specs_fixture():
                             'mount': '/lib/a',})
     _write('service', 'service-a', {'image': 'service/a'})
 
+def specs_fixture_with_depends():
+    _write('bundle', 'bundle-a', {'description': 'Bundle A', 'apps': ['appa']})
+    _write('bundle', 'bundle-b', {'description': 'Bundle B', 'apps': ['appb']})
+    _write('app', 'appa', {'repo': '/tmp/app-a',
```
thieman: Can we call the repo `repo-a` or something? Would make the tests much more clear

tests/testcases.py
```diff
@@ -136,11 +136,12 @@ def run_command(self, args, fake_exit, fake_exec_docker):
         self._clear_stdout()
         return result
 
-    def _set_up_fake_local_repo(self):
-        repo = git.Repo.init('/tmp/fake-repo')
-        with open('/tmp/fake-repo/README.md', 'w') as f:
+    def _set_up_fake_local_repo(self, path='/tmp/fake-repo'):
+        print "PATH: {}".format(path)
```
zacharym: debugging?

tests/integration/cli/restart_test.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def test_restart_nosync(self):
+        new_file_name = 'nosync_file'
+        repo = Repo.resolve(get_all_repos(include_specs_repo=False), 'app-a')
+        with open(os.path.join(repo.local_path, new_file_name), 'w+') as f:
+            f.write('new file!')
+        self.run_command('restart --no-sync appa')
+        self.assertFileNotInContainer('appa', os.path.join('/app/a/', new_file_name))
+
+    def test_restart_nosync(self):
```
thieman: Naming drep

tests/testcases.py
```diff
@@ -188,6 +189,14 @@ def assertFileContents(self, file_path, contents):
     def assertFileContentsInContainer(self, service_name, file_path, contents):
         self.assertEqual(self.exec_in_container(service_name, 'cat {}'.format(file_path)), contents)
 
+    def assertFileNotInContainer(self, service_name, file_path):
+        self.assertEqual(self.exec_in_container(service_name,
+            'sh -c \'[ -f {} ] && echo "yes" || echo "no"\''.format(file_path)).rstrip(), 'no')
+
+    def assertFileInContainer(self, service_name, file_path):
+        self.assertEqual(self.exec_in_container(service_name,
+            'sh -c \'[ -f {} ] && echo "yes" || echo "no"\''.format(file_path)).rstrip(), 'yes')
```
thieman: These should share code between each other, see some of the other examples in the class.

tests/testcases.py
```diff
@@ -208,3 +217,9 @@ def assertImageDoesNotExist(self, image):
 
     def assertExecDocker(self, *args):
         self.fake_exec_docker.assert_called_with(*args)
+
+    def inspect_container(self, service_name):
+        client = get_docker_client()
+        container_id = _get_container_for_app_or_service(client, service_name, include_exited=True)['Id']
+        return client.inspect_container(container_id)
```
thieman: Can you put this definition above the asserts for the sake of keeping some sort of structure? It's a big hairy class already.

thieman:
Design's good, bunch of nitpicks.

tests/integration/cli/restart_test.py
jsingle: It gives many milisecond digits, but on at least one occasion this method was unreliable without the sleep.  Perhaps `run_command('up')` can return before the start time per container is set?  I could look into that possibility

tests/integration/cli/restart_test.py
jsingle: Not sure what you mean.  The test cases assert that this function returns true or false

tests/integration/cli/restart_test.py
thieman: Can we do a shorter sleep?

tests/testcases.py
jsingle: Yep

jsingle:
Makin changes

tests/integration/cli/restart_test.py
jsingle: Very likely

tests/integration/cli/restart_test.py
jsingle: It would be nice to leave the functions in this file, so that the setup code can stay in this subclass as well

thieman: wat

jsingle: Talking about setting `self.up_complete_time` and the `time.sleep`

PR #302, opened 2015-06-25
are integration test broken?

zacharym:


thieman:
@paetling rsync problems

PR #309, opened 2015-06-26
docker-powered tag line

jsingle:
@thieman


thieman:
Tag Line 2: The Taggening

PR #310, opened 2015-06-26
Simplify installation page

thieman:
@jsingle This page needs to be as simple and not-scary as possible

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -20,19 +20,11 @@ brew install docker-compose
 
 ## Installation
 
-We provide a install script with each release. The install script should work
-to install Dusty from scratch, as well as to update Dusty to a new version.
-
-This install script takes the following actions:
-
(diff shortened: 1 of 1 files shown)


jsingle:
OK - I feel like somewhere we should document that our install script sets up a plist at such and such a location.  Could that go in a subsection under this copy?

thieman:
It's the standard (only?) way to define services in OS X, and the install script itself is super small and easy to reason about. I don't think we need to detail the installation at that level.

thieman:
Docker Compose didn't see a reason to https://docs.docker.com/compose/install/

jsingle:
I don't think docker-compose sets up a service for you.  There are multiple file paths that you can place your plist at.  But i agree the installation is easy to reason about

thieman:
Oh, right, derp. Point still stands, a lot of things you `apt-get` install will install services and they're not going to go out of their way to detail how that's done.

PR #313, opened 2015-06-27
Protect copy-pasters from themselves

thieman:
@jsingle Because someone will be silly enough to blindly paste this into a shell when they already have some of them installed


jsingle:
Yeah that's something that a lot of people would do.  The fix seems fine if unnecessary imo

PR #316, opened 2015-06-29
do initialize before sync

paetling:
review @jsingle


jsingle:
ah Ok.  :ghost: :camel:

PR #322, opened 2015-06-29
demote boot2docker or else bad stuff happens

jsingle:
@thieman these should for sure be demoted.  Idk if nginx and VBoxManage matter


thieman:
selfie-2

PR #326, opened 2015-06-30
add in client side option to suppress_warnings

paetling:
review @thieman   This is to solve the setup sending warnings issue.

### DIFF
--- a/dusty/daemon.py
+++ b/dusty/daemon.py
@@ -54,6 +54,8 @@ def _listen_on_socket(socket_path, suppress_warnings):
                         break
                     fn, client_version, client_suppress_warnings, args, kwargs = Payload.deserialize(data)
                     suppress_warnings |= client_suppress_warnings
+                    logging.error(suppress_warnings)
+                    logging.error(client_suppress_warnings)
                     logging.info('Received command. fn: {} args: {} kwargs: {}'.format(fn.__name__, args, kwargs))
                     try:
                         _send_warnings_to_client(connection, suppress_warnings)
--- a/dusty/daemon.py
+++ b/dusty/daemon.py
@@ -34,6 +34,9 @@ def _send_warnings_to_client(connection, suppress_warnings):
     if daemon_warnings.has_warnings and not suppress_warnings:
         connection.sendall("{}\n".format(daemon_warnings.pretty()))
 
+def _get_payload_function_data(payload):
+    return payload['fn'], payload['args'], payload['kwargs']
+
 def _listen_on_socket(socket_path, suppress_warnings):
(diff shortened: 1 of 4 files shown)


paetling:
fixing tests

dusty/payload.py
```diff
@@ -19,10 +20,11 @@ def run(self):
...
-        doc = {'fn': self.fn, 'client_version': self.client_version, 'args': self.args, 'kwargs': sorted(self.kwargs.items())}
+        doc = {'fn': self.fn, 'client_version': self.client_version, 'suppress_warnings': self.suppress_warnings,
+               'args': self.args, 'kwargs': sorted(self.kwargs.items())}
         return cPickle.dumps(doc).encode('string_escape')
 
     @staticmethod
     def deserialize(doc):
         original_doc = cPickle.loads(doc.decode('string_escape'))
-        return original_doc['fn'], original_doc['client_version'], original_doc['args'], dict(original_doc['kwargs'])
+        return original_doc['fn'], original_doc['client_version'], original_doc['suppress_warnings'], original_doc['args'], dict(original_doc['kwargs'])
```
thieman: This pattern is totally shit, can we refactor it while you're here? At least into a dict or something?

paetling: sure.

dusty/payload.py
```diff
@@ -19,10 +20,11 @@ def run(self):
         self.fn(*self.args, **self.kwargs)
 
     def serialize(self):
-        doc = {'fn': self.fn, 'client_version': self.client_version, 'args': self.args, 'kwargs': sorted(self.kwargs.items())}
```
paetling: why are we messing with kwargs here?

thieman: Deterministic behavior is generally nice. Don't think we're relying on it for anything though.

dusty/daemon.py
```diff
@@ -34,6 +34,9 @@ def _send_warnings_to_client(connection, suppress_warnings):
     if daemon_warnings.has_warnings and not suppress_warnings:
         connection.sendall("{}\n".format(daemon_warnings.pretty()))
 
+def _get_payload_function_data(payload):
+    return payload['fn'], payload['args'], payload['kwargs']
```
thieman: You dropped the `dict` cast to kwargs

thieman:
One thing otherwise :bear:

dusty/daemon.py
paetling: That should not be needed anymore because I am just passing in a strait dict to the pickle function.  Recasting what should already be a dict to a dict seems silly

thieman: ah. ok

PR #327, opened 2015-06-30
allow underscores in suite names

paetling:
review @jsingle


jsingle:
This is problematic if someone names one suite "example_suite" and another "examplesuite".  Maybe that doesn't matter, idk

paetling:
so that is problematic only if they try and run example_suite and examplesuite at the same time.  Without this change you cannot run example_suite at all.  We could also make suites have to have no udnerscores.  But that is hard to enforce and our messaging around it would be lacking. What do you think?

jsingle:
Right.  I guess I like the way this PR does it better than any alternative that I can think of

PR #328, opened 2015-06-30
Specs docs TLC

thieman:
@paetling and you thought i was just over here mindlessly banging on my keyboard

### DIFF
--- a/docs/specs/app-specs.md
+++ b/docs/specs/app-specs.md
@@ -1,121 +1,186 @@
-# Dusty App Specs
-These specs are used to specify apps for you local environment. They should go in the apps folder in your specs repo.
+# App Specs
 
-## Keys
+Apps define running containers based on source code under your control. App specs tell
+Dusty how to run this container, including any dependencies the app may have on other
+apps, services, or libs. You can also specify common scripts and testing commands for
+an app in its spec.
 
-### repo
+By default, Dusty manages the source for an app for you and keeps it up to date automatically.
+You can tell Dusty to use your own locally checked out copy of the source using `dusty repos`.
+
+## repo
+
 ```
-repo: github.com/gamechanger/app1
-repo: /Users/dusty/app1
+repo: github.com/my-org/my-app
+  -or-
+repo: /Users/myuser/my-app
 ```
-This repo key is used to specify a local or remote GitHub repo where the code for the app lives.
 
-### mount
+`repo` specifies the repo containing the source for an app. By default, Dusty manages this
+repo for you and will keep its local copy up to date. Once a repo is defined in an active spec,
+it can be controlled using the `dusty repos` command.
+
+Repos can be specified using either a URL or an absolute path to a Git repo on your local filesystem.
+
+## mount
+
 ```
-mount: /gc/app1
+mount: /my-app
 ```
-The mount key is used to specify where in the container's file system to mount the code specified by the repo key.  The mounting is done using Docker's volume mounts. <br />
-Using the image and mount keys specified above, the code specified in `/Users/dusty/app1` on your Mac will be visible in your container at `/gc/app1`.<br />
-The repo and mount keys are linked.  If you specify one you must specify the other.<br />
 
+`mount` tells Dusty where to mount the contents of the repo inside the running container.
 
-### depends
+`mount` must be provided if `repo` is provided.
+
+## depends
+
 ```
 depends:
   services:
-    - coreMongo
-    - coreRedis
+    - myService1
+    - myService2
   apps:
-    - deferrableworker
-    - docketsworker
+    - myOtherApp
   libs:
-    - gclib
+    - myLib
 ```
-The depends key is used to specify what libs, apps and services this app depends on.<br />
-Apps and services specified in the depends dict are spun up before the current app. The current app is then linked (using Docker Compose's links) to the apps and services. Because we are using Docker Compose links, dependencies must only be 1 way.  This means you cannot have app1 depend on app2 and app2 depend on app1.
 
-Libs specified in the depends dict have a different meaning.  They tell Dusty that these libs (and their recursive dependencies) need to be mounted into this app's container (using Docker's volume mounts).  In so doing, the running app will have access to the most recent code in these libs.
+`depends` is used to specify what libs, apps and services this app depends on.
 
-### conditional_links
+Containers for apps and services specified in the `depends` dict are created before the container
+for the referencing app. The referencing app is then linked (using [Docker links](https://docs.docker.com/userguide/dockerlinks/))
+to the dependent apps and services. Once containers are linked, they may reference each other through
+the environment variables or `/etc/hosts` overrides provided by Docker links.
+
+**N.B.:** All container links, including those from both `depends` and `conditional_links`, must be acyclical.
+
+Dependent libs (and their lib dependencies) will be mounted into the running app container and
+installed according to their specs.
+
+## conditional_links
+
 ```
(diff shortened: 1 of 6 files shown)


docs/specs/app-specs.md
```diff
@@ -1,121 +1,186 @@
...
+# App Specs
 
-## Keys
+Apps define running containers based on source code under your control. App specs tell
+Dusty how to run this container, including any dependencies the app may have on other
+apps, services, or libs. You can also specify common scripts and testing commands for
+an app in its spec.
+
+By default, Dusty manages the source for an app for you and keeps it up to date automatically.
+You can tell Dusty to use your own locally checked out copy of the source using `dusty repos`.
```
paetling: this is not app specific stuff so should probably go in index

thieman: I think it's more important that docs show people what they're looking for than that they're DRY. If we were using a more robust docs system we could have fact tables and stuff but we don't, do we just end up copy-pasting.

Do you not think this is useful here to define what an app is?

docs/specs/app-specs.md
```diff
@@ -1,121 +1,186 @@
...
 
-Libs specified in the depends dict have a different meaning.  They tell Dusty that these libs (and their recursive dependencies) need to be mounted into this app's container (using Docker's volume mounts).  In so doing, the running app will have access to the most recent code in these libs.
+`depends` is used to specify what libs, apps and services this app depends on.
+
+Containers for apps and services specified in the `depends` dict are created before the container
+for the referencing app. The referencing app is then linked (using [Docker links](https://docs.docker.com/userguide/dockerlinks/))
+to the dependent apps and services. Once containers are linked, they may reference each other through
+the environment variables or `/etc/hosts` overrides provided by Docker links.
+
+**N.B.:** All container links, including those from both `depends` and `conditional_links`, must be acyclical.
```
paetling: what does N.B. stand for?

docs/specs/app-specs.md
```diff
@@ -1,121 +1,186 @@
...
-2. It is specified in an activated bundles apps list. <br/>
-As in the depends dict, circular dependencies are not allowed.
 
-### host_forwarding
+`conditional_links` specifies downstream apps which are not required by the referencing app but
+should be linked if they are running as the result of a different bundle or dependency. This lets
+you layer optional containers onto a stack without forcing them all to be run whenever the main
+container is run.
+
+Unlike `depends`, `conditional_links` do not support links to libs.
```
paetling: They do not support services either right?

docs/specs/app-specs.md
```diff
@@ -1,121 +1,186 @@
...
     container_port: 80
 ```
-The host_forwarding key is used specify a list of dictionaries.  Each dictionary allows you to specify a mapping from the containers ip space to your Mac's ip space.<br />
-**host_name** is the local url you would like to hit.<br />
-**host_port** is the port on which you will hit the host_name.<br />
-**container_port** is the port the app is running on in the Docker container.<br />
 
-### build
+`host_forwarding` allows you to specify multiple routes which should be linked between your host OS
+and the application running inside this app's container.
```
paetling: when I read routes i do not think urls.  might be easier just to say urls?

docs/specs/app-specs.md
thieman: They should, and schema suggests they do.

docs/specs/app-specs.md
thieman: Latin for *pay attention* https://en.wikipedia.org/wiki/Nota_bene

docs/specs/index.md
```diff
@@ -1,22 +1,37 @@
...
+1. [Apps](./app-specs.md): Applications which you actively develop. These may link
+to a source repo and install Libs. May include test specifications.
+1. [Services](./service-specs.md): Applications which you do not develop. Services are
+often used to run database containers off of a public Docker image.
+1. [Libs](./lib-specs.md): Libraries which you actively develop. These may be depended on
+by Apps, which will then install them automatically. May include test specifications.
+
+Additionally, apps and libs share the following common sub-schemata:
+
+1. [Tests](./test-specs.md): Specifications of test images and suites. Libs and Apps with
```
paetling: This is showing up as 1. might want it to actually be 5?

docs/specs/index.md
```diff
@@ -1,22 +1,37 @@
...
-There is a 5th subtype that is important enough to get its own spec page.<br/>
-5. [Tests](./test-specs.md) - objects found on both the lib and apps objects. They define how to run tests on a user's app and lib code.
+Dusty uses YAML specifications for configuration of the applications, services, and
+libraries used to create your stack.
+
+An additional entrypoint layer, called a Bundle, is used to provide toggleable entrypoints
+into your stack's dependency graph. Users can decide which Bundles they want to run,
+and Dusty runs the applications, services, and libraries defined by those Bundles.
+
+1. [Bundles](./bundle-specs.md): User-toggleable logical groups of applications.
```
paetling: User-toggleable logical groups is a mouth full.   Logical groups of applications users can toggle?

docs/specs/index.md
thieman: I actually don't think that's possible with this implementation of markdown

docs/specs/lib-specs.md
```diff
@@ -1,30 +1,69 @@
...
 ```
 install:
-  - python setup.py install --force
+  - python setup.py install
 ```
-The install key is used to specify a list of commands that should be run to prepare the library to be used. This command is run in the app's container, prior to the apps once and always commands being run.
 
-### depends
+`install` specifies a list of commands that should be run to prepare the library to be used.
+This command is run the container of an app which depends on the lib.
```
paetling: run in the container*

docs/specs/test-specs.md
```diff
@@ -1,39 +1,114 @@
...
 
-## Keys
+## image
+
+```
+image: ubuntu:15.04
+```
+
+`image` specifies a Docker image on which to base the container which will be used
+to run the test suite. If the image does not exist locally, Dusty will pull it.
```
paetling: this is not true.  Image is used to build a second image which will be used to build the container.

paetling:
No where do we talk about the ability to run commands in the background. That is probably worth mentioning.

docs/specs/app-specs.md
paetling: When I first read this i thought you were taking about dusty managing dusty-specs, which I think is not worth mentioning in apps. Reading it now it seems more like you are talking about repos stuff, which is worth mentioning

docs/specs/app-specs.md
paetling: you are right

docs/specs/index.md
paetling: Then maybe it is not worth numbering because there is only one anyway?

docs/specs/app-specs.md
thieman: URLs have schemes, so that's inaccurate. Really it's exactly what the spec says, a hostname and a port. Not sure what the best noun is for that.

docs/specs/index.md
thieman: Sure, that looks better

docs/specs/app-specs.md
paetling: yeah.  i think since you explain below this is fine.

paetling:
some thoughts and suggestions, but looks good :ghost: :cow:

thieman:
feedback committed, margin

PR #330, opened 2015-06-30
Refresh warnings

jsingle:
@paetling

### DIFF
--- a/dusty/daemon.py
+++ b/dusty/daemon.py
@@ -15,7 +15,7 @@ import socket
 
 from docopt import docopt
 
-from .preflight import preflight_check
+from .preflight import preflight_check, refresh_preflight_warnings
 from .log import configure_logging, make_socket_logger, close_socket_logger
 from .constants import SOCKET_PATH, SOCKET_TERMINATOR, SOCKET_ERROR_TERMINATOR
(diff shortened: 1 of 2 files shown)


dusty/daemon.py
```diff
@@ -61,6 +66,7 @@ def _listen_on_socket(socket_path, suppress_warnings):
                     suppress_warnings |= payload['suppress_warnings']
                     logging.info('Received command. fn: {} args: {} kwargs: {}'.format(fn.__name__, args, kwargs))
                     try:
+                        _refresh_warnings()
```
paetling: why are you doing this?

jsingle: Well warnings never get changed after preflight, even if someone upgrades the version of one of boot2docker, nginx, etc.  which is confusing

paetling: if you update something installed and then want daemon to automatically know about it?

jsingle: Don't want it to keep saying "WARNING yo shit is out of date" when it's no longer out of date

paetling: yep. We are on the same page.

paetling:
:ghost: :cow:

PR #332, opened 2015-06-30
Js send json

jsingle:
@thieman this should fix the security vulnerability.  I like this system of just using a decorator to register a function, instead of typing out a big map somewhere

### DIFF
--- a/dusty/payload.py
+++ b/dusty/payload.py
@@ -43,7 +43,7 @@ def daemon_command(f):
     _daemon_command_mapping[key] = f
     return f
 
-def run_command_as_daemon(fn_key):
+def get_payload_function(fn_key):
     if fn_key not in _daemon_command_mapping:
         raise RuntimeError('Function key {} not found'.format(fn_key))
-    _daemon_command_mapping[fn_key]()
+    _daemon_command_mapping[fn_key]
--- a/dusty/payload.py
+++ b/dusty/payload.py
@@ -1,4 +1,4 @@
-import cPickle
+import json
 
 from .constants import VERSION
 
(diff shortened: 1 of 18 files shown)


tests/integration/cli/config_test.py
```diff
@@ -27,4 +27,4 @@ def test_config_set_fails_on_nonexistent_key(self):
     def test_config_set_works_with_valid_input(self):
         self.run_command('config set nginx_includes_dir /var/nginx')
         result = yaml.load(self.run_command('config listvalues'))
-        self.assertEqual(result['nginx_includes_dir'], '/var/nginx')
+        self.assertEqual(result['nginx_includes_dir'], u'/var/nginx')
```
thieman: This shouldn't be necessary, was it failing without it?

dusty/payload.py
```diff
@@ -20,11 +20,30 @@ def run(self):
...
+            raise RuntimeError('Function key {} not found; you may need to decorate your function'.format(fn_key))
+        doc = {'fn_key': fn_key, 'client_version': self.client_version, 'suppress_warnings': self.suppress_warnings,
                'args': self.args, 'kwargs': self.kwargs}
-        return cPickle.dumps(doc).encode('string_escape')
+        return json.dumps(doc).encode('string_escape')
 
     @staticmethod
     def deserialize(doc):
-        return cPickle.loads(doc.decode('string_escape'))
+        return json.loads(doc.decode('string_escape'))
```
thieman: Do we still need `string_escape` if we're doing JSON?

dusty/payload.py
```diff
@@ -20,11 +20,30 @@ def run(self):
...
 
+_daemon_command_mapping = {}
+
+def _get_key(f):
+    return '{}.{}'.format(f.__module__, f.__name__)
```
thieman: This function needs a better name

dusty/payload.py
```diff
@@ -20,11 +20,30 @@ def run(self):
...
+
+def daemon_command(f):
+    key = _get_key(f)
+    if key in _daemon_command_mapping and _daemon_command_mapping[key] != f:
+        raise RuntimeError("Function mapping key collision: {}. Name one of the functions something else".format(key))
+    _daemon_command_mapping[key] = f
+    return f
```
thieman: `fn` would be a little better for comprehension than `f`

tests/integration/cli/config_test.py
jsingle: Yeah, and with it.  The issue is that json loads makes all the strings unicode, even if you dumped non-unicode strings.  Apparently using yaml.safe_load gets around this (but won't work the same for weird character sets)

thieman:
Like the design, couple nitpicks otherwise :bear:

jsingle:
Still gotta look into string escape.  I wish I understood how strings work :/

PR #333, opened 2015-06-30
add ability to set how much memory you would like to run with and change memory on up

paetling:
review @jsingle

### DIFF
--- a/dusty/commands/setup.py
+++ b/dusty/commands/setup.py
@@ -66,14 +66,12 @@ def _get_and_configure_nginx_includes_dir():
     return ''
 
 def _get_boot2docker_vm_size():
-    memory_gigs = 1.0 * virtual_memory().total / 10**9
-    if memory_gigs >= 15:
+    memory_gigs = virtual_memory().total / 2**30
+    if memory_gigs == 16:
         vm_gigs = 6
-        host_gigs = 16
     else:
         vm_gigs = 4
-        host_gigs = 8
-    if _get_raw_input('Your system seems to have {} gigabytes of memory. We would like to allocate {} to your vm. Is that ok? (y/n) '.format(host_gigs, vm_gigs)).upper() == 'Y':
+    if _get_raw_input('Your system seems to have {} gigabytes of memory. We would like to allocate {} to your vm. Is that ok? (y/n) '.format(memory_gigs, vm_gigs)).upper() == 'Y':
         return vm_gigs
     else:
         return int(_get_raw_input('Please input the number of gigabytes to allocate to the vm: '))
--- a/dusty/commands/setup.py
+++ b/dusty/commands/setup.py
@@ -66,15 +66,19 @@ def _get_and_configure_nginx_includes_dir():
     return ''
 
 def _get_boot2docker_vm_size():
-    memory_gigs = virtual_memory().total / 2**30
-    if memory_gigs == 16:
-        vm_gigs = 6
+    memory_megs = virtual_memory().total / 2**20
+    sixteen_megs = 16 * 2**10
+    eight_megs = 8 * 2**10
+    if memory_megs >= sixteen_megs:
+        vm_megs = 6 * 2**10
+    elif memory_megs >= eight_megs:
+        vm_megs = 4 * 2**10
     else:
-        vm_gigs = 4
-    if _get_raw_input('Your system seems to have {} gigabytes of memory. We would like to allocate {} to your vm. Is that ok? (y/n) '.format(memory_gigs, vm_gigs)).upper() == 'Y':
-        return vm_gigs
+        vm_megs = 2 * 2**10
+    if _get_raw_input('Your system seems to have {} megabytes of memory. We would like to allocate {} to your vm. Is that ok? (y/n) '.format(memory_megs, vm_megs)).upper() == 'Y':
+        return vm_megs
     else:
-        return int(_get_raw_input('Please input the number of gigabytes to allocate to the vm: '))
+        return int(_get_raw_input('Please input the number of megabytes to allocate to the vm: '))
 
 def setup_dusty_config(mac_username=None, specs_repo=None, nginx_includes_dir=None):
     print "We just need to verify a few settings before we get started.\n"
--- a/dusty/commands/setup.py
+++ b/dusty/commands/setup.py
@@ -3,6 +3,7 @@ import subprocess
 import textwrap
 from os.path import isfile, isdir
 from os import mkdir
+from psutil import virtual_memory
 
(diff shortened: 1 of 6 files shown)


dusty/commands/setup.py
```diff
@@ -64,6 +65,19 @@ def _get_and_configure_nginx_includes_dir():
     _get_raw_input('\n'.join(textwrap.wrap('You have a custom nginx config setup. Could not find an nginx.conf file. Please read our docs to see what is needed for the nginx config.  Once you have figured it out, please use `dusty config` command to adjust your `nginx_includes_dir`', 80)))
     return ''
 
+def _get_boot2docker_vm_size():
+    memory_gigs = 1.0 * virtual_memory().total / 10**9
```
jsingle: divide by `2**30` not `10**9`

dusty/commands/setup.py
```diff
@@ -64,6 +65,19 @@ def _get_and_configure_nginx_includes_dir():
...
+    if memory_gigs >= 15:
+        vm_gigs = 6
+        host_gigs = 16
+    else:
+        vm_gigs = 4
+        host_gigs = 8
+    if _get_raw_input('Your system seems to have {} gigabytes of memory. We would like to allocate {} to your vm. Is that ok? (y/n) '.format(host_gigs, vm_gigs)).upper() == 'Y':
```
jsingle: Can we not use the variable `host_gigs` and just stick with one measure of `memory_gigs`?  Arbitrarily setting `host_gigs` seems silly

thieman:
VirtualBox does everything in MB, so we should too.

dusty/commands/setup.py
```diff
@@ -64,6 +65,17 @@ def _get_and_configure_nginx_includes_dir():
     _get_raw_input('\n'.join(textwrap.wrap('You have a custom nginx config setup. Could not find an nginx.conf file. Please read our docs to see what is needed for the nginx config.  Once you have figured it out, please use `dusty config` command to adjust your `nginx_includes_dir`', 80)))
     return ''
 
+def _get_boot2docker_vm_size():
+    memory_gigs = virtual_memory().total / 2**30
+    if memory_gigs == 16:
+        vm_gigs = 6
+    else:
+        vm_gigs = 4
```
thieman: Should probably decay to 2GB if the system has less than 8 gigs total

dusty/commands/setup.py
```diff
@@ -64,6 +65,17 @@ def _get_and_configure_nginx_includes_dir():
     _get_raw_input('\n'.join(textwrap.wrap('You have a custom nginx config setup. Could not find an nginx.conf file. Please read our docs to see what is needed for the nginx config.  Once you have figured it out, please use `dusty config` command to adjust your `nginx_includes_dir`', 80)))
     return ''
 
+def _get_boot2docker_vm_size():
+    memory_gigs = virtual_memory().total / 2**30
+    if memory_gigs == 16:
```
thieman: `>=`?

dusty/commands/setup.py
```diff
@@ -64,6 +65,21 @@ def _get_and_configure_nginx_includes_dir():
     _get_raw_input('\n'.join(textwrap.wrap('You have a custom nginx config setup. Could not find an nginx.conf file. Please read our docs to see what is needed for the nginx config.  Once you have figured it out, please use `dusty config` command to adjust your `nginx_includes_dir`', 80)))
     return ''
 
+def _get_boot2docker_vm_size():
+    memory_megs = virtual_memory().total / 2**20
+    sixteen_megs = 16 * 2**10
```
jsingle: this is weird naming.  This is actually 16 gigs, expresses as a number of megabytes, right?

dusty/config.py
```diff
@@ -27,7 +27,8 @@ def write_default_config():
     default_config = {constants.CONFIG_BUNDLES_KEY: [],
                       constants.CONFIG_REPO_OVERRIDES_KEY: {},
                       constants.CONFIG_NGINX_DIR_KEY: '/usr/local/etc/nginx/servers',
-                      constants.CONFIG_SETUP_KEY: False}
+                      constants.CONFIG_SETUP_KEY: False,
+                      constants.CONFIG_VM_MEM_SIZE: '2'}
```
jsingle: shouldn't this be 2048?

dusty/constants.py
```diff
@@ -65,7 +66,8 @@
     CONFIG_MAC_USERNAME_KEY: 'The user on the host OS who will own and be able to access the boot2docker VM. Dusty runs all VirtualBox, boot2docker, Docker, and Docker Compose commands as this user.',
     CONFIG_SPECS_REPO_KEY: 'This repository is used for storing the specs used by Dusty.  It is managed the same way as other repos',
     CONFIG_NGINX_DIR_KEY: 'This is the location that your nginx config will import extra files from.  Dusty\'s nginx config will be stored here',
-    CONFIG_SETUP_KEY: 'Key indicating if you have run the required command `dusty setup`'
+    CONFIG_SETUP_KEY: 'Key indicating if you have run the required command `dusty setup`',
+    CONFIG_VM_MEM_SIZE: 'Specifies how much memory (in gigabytes) you want your boot2docker vm to have'
```
jsingle: Are we using GB or MB?  It seems like the function returns MB

dusty/commands/setup.py
paetling: that is a good point.   Will adjust

dusty/config.py
paetling: no.   based of a return type passed in. Maybe for clarity it should be.  Will change it.

jsingle:
Looks good

dusty/commands/setup.py
paetling: I changed my mind.  This is a bad point.  A megabyte is 2 ^ 20.   So what sixteen_megs here is the total bytes for 16 megs.

dusty/commands/setup.py
```diff
@@ -64,6 +65,20 @@ def _get_and_configure_nginx_includes_dir():
...
 
+def _get_boot2docker_vm_size():
+    # all math is done in megabytes
+    memory_megs = virtual_memory().total / 2**20
+    if memory_megs >= 16 * 2**10:
+        vm_megs = 6 * 2**10
+    elif memory_megs >= 8 * 2**10:
+        vm_megs = 4 * 2**10
+    else:
+        vm_megs = 2 * 2**10
```
thieman: Generally when you're doing math like this you want to create unit constants to make it easier to read and harder to make mistakes, e.g.

```
GIGABYTES = 2**10

vm_megs = 6 * GIGABYTES
```

PR #335, opened 2015-07-01
add ssh sock get on every dusty command

paetling:
review @thieman  is this sort of what you had in mind?

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -4,7 +4,7 @@
   * NEW: Using a config variable, Dusty now manages how large (in megabytes) to make your vm.  On each start of boot2docker, Dusty will adjust your vm's memory size.
 
   * FIXED: Install script now passes -H flag to `sudo dusty -d --preflight-only`, so that `boot2docker version` won't create `~/.boot2docker` owned by root.
-  * FIXED: We are now loading the SSH_AUTH_SOCK with each Dusty command. This fixes a race condition in launchd between Dusty and SSH_AUTH_SOCK.
+  * FIXED: We are now attempting to load the SSH_AUTH_SOCK, if it is not currently set, with each Dusty command. This fixes a race condition in launchd between Dusty and SSH_AUTH_SOCK.
 
 ## 0.1.4 (June 29, 2015)
   * BREAKING CHANGE: `services` and `compose` keys in test specs are now specified at the suite level, not the test level.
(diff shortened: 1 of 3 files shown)


docs/changelog.md
```diff
@@ -4,6 +4,7 @@
   * NEW: Using a config variable, Dusty now manages how large (in megabytes) to make your vm.  On each start of boot2docker, Dusty will adjust your vm's memory size.
 
   * FIXED: Install script now passes -H flag to `sudo dusty -d --preflight-only`, so that `boot2docker version` won't create `~/.boot2docker` owned by root.
+  * FIXED: The race condition in launchd between Dusty and SSH_AUTH_SOCK has been removed.
```
thieman: Actually say what you changed, then say what it fixes.

dusty/daemon.py
```diff
@@ -42,6 +42,14 @@ def _refresh_warnings():
         refresh_config_warnings()
         refresh_preflight_warnings()
 
+def _run_pre_command_functions(connection, suppress_warnings, client_version):
+    _refresh_warnings()
+    _send_warnings_to_client(connection, suppress_warnings)
+    if client_version != VERSION:
+        raise RuntimeError("Dusty daemon is running version: {}, and client is running version: {}".format(VERSION, client_version))
```
thieman: Version check should be done before any of the other stuff, and it should probably be separate from the rest of the functionality we'll stuff in this function

dusty/daemon.py
```diff
@@ -42,6 +42,12 @@ def _refresh_warnings():
         refresh_config_warnings()
         refresh_preflight_warnings()
 
+def _run_pre_command_functions(connection, suppress_warnings, client_version):
+    _refresh_warnings()
+    _send_warnings_to_client(connection, suppress_warnings)
+    check_and_load_ssh_auth()
```
thieman: We should probably only do this if we've failed to get the socket previously. Also, refresh warnings should be the *last* thing we do in this block. Successfully acquiring the socket should clear the warning.

docs/changelog.md
```diff
@@ -4,6 +4,7 @@
   * NEW: Using a config variable, Dusty now manages how large (in megabytes) to make your vm.  On each start of boot2docker, Dusty will adjust your vm's memory size.
 
   * FIXED: Install script now passes -H flag to `sudo dusty -d --preflight-only`, so that `boot2docker version` won't create `~/.boot2docker` owned by root.
+  * FIXED: We are now loading the SSH_AUTH_SOCK with each Dusty command. This fixes a race condition in launchd between Dusty and SSH_AUTH_SOCK.
```
thieman: Remember to correct this if you address my other comment

dusty/daemon.py
```diff
@@ -42,6 +42,12 @@ def _refresh_warnings():
         refresh_config_warnings()
         refresh_preflight_warnings()
 
+def _run_pre_command_functions(connection, suppress_warnings, client_version):
+    _refresh_warnings()
+    _send_warnings_to_client(connection, suppress_warnings)
```
thieman: Should this line also not be grouped with this function? I feel like we have different ideas of what actually belongs here

PR #342, opened 2015-07-02
Ae container life cycles

paetling:
review @jsingle

### DIFF
--- a/dusty/commands/test.py
+++ b/dusty/commands/test.py
@@ -163,7 +163,7 @@ def _run_tests_with_image(client, expanded_specs, app_or_lib_name, test_command,
     client.remove_container(container=test_container_name)
 
     for service_container in previous_container_names:
-        log_to_client('Stopping service container {}'.format(service_container))
+        log_to_client('Killing service container {}'.format(service_container))
         client.kill(service_container)
         client.remove_container(container=service_container)
     return exit_code
--- a/dusty/commands/test.py
+++ b/dusty/commands/test.py
@@ -166,6 +166,8 @@ def _run_tests_with_image(client, expanded_specs, app_or_lib_name, test_command,
         log_to_client('Stopping service container {}'.format(service_container))
         try:
             client.stop(service_container, timeout=1)
+            client.wait(service_container)
+            client.remove_container(container=service_container)
         except Exception as e:
             log_to_client('Exception stopping service container {}: {}'.format(service_container, str(e)))
     return exit_code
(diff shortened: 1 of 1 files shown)


dusty/commands/test.py
```diff
@@ -143,10 +143,8 @@ def _run_tests_with_image(client, expanded_specs, app_or_lib_name, test_command,
 
     for service_container in previous_container_names:
         log_to_client('Stopping service container {}'.format(service_container))
-        try:
-            client.stop(service_container, timeout=1)
-        except Exception as e:
-            log_to_client('Exception stopping service container {}: {}'.format(service_container, str(e)))
+        client.kill(service_container)
+        client.remove_container(container=service_container)
```
thieman: Struggling to remember why we didn't do this in the first place. Maybe it was because we were thinking we'd share service containers across suites then?

dusty/commands/test.py
```diff
@@ -143,10 +143,8 @@ def _run_tests_with_image(client, expanded_specs, app_or_lib_name, test_command,
 
     for service_container in previous_container_names:
         log_to_client('Stopping service container {}'.format(service_container))
```
thieman: Should also change this to killing

dusty/commands/test.py
```diff
@@ -143,10 +143,8 @@ def _run_tests_with_image(client, expanded_specs, app_or_lib_name, test_command,
...
-        try:
-            client.stop(service_container, timeout=1)
-        except Exception as e:
-            log_to_client('Exception stopping service container {}: {}'.format(service_container, str(e)))
+        client.kill(service_container)
```
thieman: Actually why do we need to do this instead of the stop with a short timeout to give them some chance to die with dignity?

paetling: Why do we care if they die with dignity if we are just going to remove all traces of this container next.   Dieing with dignity is reserver for containers we respect. We do not care or respect this container.

thieman: This is making an assumption that the containers won't persist anything, which technically we don't know.

jsingle:
Seems good to me

dusty/commands/test.py
paetling: How would you persist something using a testing service container?   You would have to have compose spec set volumes to a mounted volume right?   This seems like a use case we do not want to encourage.  We want are trying to isolate these test runs right?

thieman: Everything you're saying is correct. Your decision.

PR #344, opened 2015-07-02
add --all command

paetling:
review @theiman

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## 0.1.6 (In Progress)
-  * NEW: Dusty now offers a --all option when running tests.  This will run all test suites associated with an app or lib.
+  * NEW: Dusty now offers a --all option when running tests.  This will run all test suites associated with an app or lib. Also, `all` is now a reserved suite_name.
 
   * When creating test images, logs are now streamed to the user.
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## 0.1.6 (In Progress)
+  * NEW: Dusty now offers a --all option when running tests.  This will run all test suites associated with an app or lib.
+
   * When creating test images, logs are now streamed to the user.
 
 ## 0.1.5 (July 1, 2015)
--- a/dusty/cli/test.py
+++ b/dusty/cli/test.py
@@ -11,7 +11,6 @@ Options:
   --no-pull     Do not pull dusty managed repos from remotes.
   --all         Run all test suites
 
-
 Examples:
   To call test suite frontend with default arguments:
     dusty test web frontend
--- a/dusty/cli/test.py
+++ b/dusty/cli/test.py
@@ -27,8 +27,6 @@ from ..commands.test import (run_app_or_lib_tests, test_info_for_app_or_lib, pul
 
 def main(argv):
     args = docopt(__doc__, argv, options_first=True)
-    import logging
-    logging.error(args)
     if args['--all']:
         payload0 = Payload(pull_repos_and_sync_commands,
                            args['<app_or_lib_name>'],
--- a/dusty/cli/test.py
+++ b/dusty/cli/test.py
@@ -2,14 +2,14 @@
 If args are passed, default arguments are dropped
 
 Usage:
-  test [options] <app_or_lib_name> [--all] [<suite_name>] [<args>...]
+  test [options] <app_or_lib_name> [<suite_name>] [<args>...]
 
 Options:
   <suite_name>  Name of the test suite you would like to run
+                This can also be --all to run all suites in the spec
   <args>        A list of arguments to be passed to the test script
   --recreate    Ensures that the testing image will be recreated
   --no-pull     Do not pull dusty managed repos from remotes.
-  --all         Run all test suites
 
 Examples:
   To call test suite frontend with default arguments:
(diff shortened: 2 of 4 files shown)


docs/changelog.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 0.1.6 (In Progress)
+  * NEW: Dusty now offers a --all option when running tests.  This will run all test suites associated with an app or lib.
```
thieman: offers an

dusty/cli/test.py
```diff
@@ -22,13 +23,13 @@
...
 from ..commands.test import (run_app_or_lib_tests, test_info_for_app_or_lib, pull_repos_and_sync_commands,
-                             ensure_valid_suite_name)
+                             ensure_valid_suite_name, run_all_app_or_lib_suites)
 
 def main(argv):
     args = docopt(__doc__, argv, options_first=True)
     if not args['<suite_name>']:
         return Payload(test_info_for_app_or_lib, args['<app_or_lib_name>'])
-    else:
+    elif not args['--all']:
```
thieman: Nitpick but generally it's easier to understand branches if you avoid negation whenever possible. So this could be `elif args['--all']` and switched with the else branch.

dusty/cli/test.py
```diff
@@ -2,13 +2,14 @@
 If args are passed, default arguments are dropped
 
 Usage:
-  test [options] <app_or_lib_name> [<suite_name>] [<args>...] [--recreate] [--no-pull]
+  test [options] <app_or_lib_name> [--all] [<suite_name>] [<args>...]
```
thieman: Does docopt give us a way to say you can either provide --all or you can provide a suite_name with args?

thieman:
Good on design

dusty/cli/test.py
jsingle: I believe it does

docs/changelog.md
paetling: why?  you say it dash dash all option, right?

dusty/cli/test.py
paetling: I agree, but was trying to follow the same patter as above.

dusty/cli/test.py
paetling: will look into this

docs/changelog.md
thieman: My internal monologue just sadi "offers a all option" but whatever, trivial point :stuck_out_tongue:

dusty/cli/test.py
```diff
@@ -21,16 +22,23 @@
...
-                             ensure_valid_suite_name)
+from ..commands.test import (run_app_or_lib_tests, test_info_for_app_or_lib, pull_repos_and_sync,
+                             ensure_valid_suite_name, run_all_app_or_lib_suites)
 
 def main(argv):
     args = docopt(__doc__, argv, options_first=True)
-    if not args['<suite_name>']:
-        return Payload(test_info_for_app_or_lib, args['<app_or_lib_name>'])
-    else:
+    if args['<suite_name>'] == '--all':
```
thieman: This is crazy hacky, if something has `--` in front of it it should definitely be an option, not an arg. If we're going to do it this way I think it'd be better to just make `all` a reserved suite name.

paetling: Yeah this is why i wanted you to take a look.  Docopt refuses to distinguish --all option I am trying to provide.  If i put it after gcweb it will always interpret it as <suite_name>. I like the idea of reserving all.

dusty/commands/test.py
```diff
@@ -60,15 +61,30 @@ def pull_repos_and_sync_commands(app_or_lib_name, pull_repos=False):
     make_test_command_files(app_or_lib_name, expanded_specs)
     if pull_repos:
         _update_test_repos(app_or_lib_name)
+    spec = expanded_specs.get_app_or_lib(app_or_lib_name)
+    sync_repos_by_specs([spec])
 
-def run_app_or_lib_tests(app_or_lib_name, suite_name, test_arguments, force_recreate=False):
+def run_app_or_lib_tests(app_or_lib_name, suite_name, test_arguments, should_exit=True, force_recreate=False):
```
thieman: When is should_exit false?

dusty/commands/test.py
```diff
@@ -60,15 +61,30 @@ def pull_repos_and_sync_commands(app_or_lib_name, pull_repos=False):
...
     ensure_test_image(client, app_or_lib_name, expanded_specs, force_recreate=force_recreate)
-    _run_tests_with_image(client, expanded_specs, app_or_lib_name, test_command, suite_name)
+    return _run_tests_with_image(client, expanded_specs, app_or_lib_name, test_command, suite_name, should_exit=should_exit)
+
+def run_all_app_or_lib_suites(app_or_lib_name, force_recreate=False):
+    expanded_specs = get_expanded_libs_specs()
+    spec = expanded_specs.get_app_or_lib(app_or_lib_name)
+    exit_code = 0
+    for index, suite_spec in enumerate(spec['test']['suites']):
+        args = [app_or_lib_name, suite_spec['name'], [], False]
```
thieman: This is really hard to parse, I think it'd be more clear if `should_exit` and `force_recreate` were handled as kwargs. Also less brittle.

paetling: fair enough.  will make that change.

dusty/commands/test.py
```diff
@@ -148,5 +164,7 @@ def _run_tests_with_image(client, expanded_specs, app_or_lib_name, test_command,
         except Exception as e:
             log_to_client('Exception stopping service container {}: {}'.format(service_container, str(e)))
 
-    log_to_client('TESTS {}'.format('FAILED' if exit_code != 0 else 'PASSED'))
-    sys.exit(exit_code)
+    if should_exit:
+        log_to_client('TESTS {}'.format('FAILED' if exit_code != 0 else 'PASSED'))
+        sys.exit(exit_code)
```
thieman: Really this logic shouldn't even be in here, the callers should do it.

PR #345, opened 2015-07-06
Js upgrade

jsingle:
@thieman

### DIFF
--- a/dusty/cli/upgrade.py
+++ b/dusty/cli/upgrade.py
@@ -21,5 +21,5 @@ from ..commands.upgrade import upgrade_dusty_binary
 def main(argv):
     args = docopt(__doc__, argv)
     if args['<version>']:
-      return Payload(upgrade_dusty_binary, args['<version>'])
+        return Payload(upgrade_dusty_binary, args['<version>'])
     return Payload(upgrade_dusty_binary)
--- a/dusty/cli/upgrade.py
+++ b/dusty/cli/upgrade.py
@@ -1,16 +1,16 @@
 """Upgrade Dusty's binaries
 
-If Dusty is being run with a binary, this command will replace the
-binary with the specified version from GitHub.  If no version is
-specified, the latest version of the binary will be used.  The
-daemon will then exec into the new version of the binary.
+Upgrades Dusty to the specified version.  If no version is
+specified, this will upgrade to the latest version.  This command
+only works if Dusty is being run as a binary (as opposed to running
+from source).
 
 Usage:
   upgrade [<version>]
 
 Options:
-  <version>         If provided, this version of Dusty will be downloaded
-                  and used (defaults to use the most recent version)
+  <version>     If provided, this version of Dusty will be downloaded
+                and used (defaults to use the most recent version)
 """
 
 from docopt import docopt
--- a/dusty/cli/upgrade.py
+++ b/dusty/cli/upgrade.py
@@ -1,0 +1,25 @@
+"""Upgrade Dusty's binaries
+
+If Dusty is being run with a binary, this command will replace the
+binary with the specified version from GitHub.  If no version is
+specified, the latest version of the binary will be used.  The
+daemon will then exec into the new version of the binary.
+
+Usage:
+  upgrade [<version>]
+
+Options:
+  <version>         If provided, this version of Dusty will be downloaded
+                  and used (defaults to use the most recent version)
+"""
+
+from docopt import docopt
+
+from ..payload import Payload
+from ..commands.upgrade import upgrade_dusty_binary
+
+def main(argv):
+    args = docopt(__doc__, argv)
+    if args['<version>']:
+      return Payload(upgrade_dusty_binary, args['<version>'])
+    return Payload(upgrade_dusty_binary)
--- a/dusty/commands/upgrade.py
+++ b/dusty/commands/upgrade.py
@@ -59,11 +59,15 @@ def _move_temp_binary_to_path():
 @daemon_command
 def upgrade_dusty_binary(version=None):
     if not constants.BINARY:
-        log_to_client("The dusty upgrade command only works when dusty is run with its binaries")
+        log_to_client('It looks like you\'re running Dusty from source. Upgrade is only available when you use the installed binary')
         return
     if version is None:
         version = _get_latest_version()
-        log_to_client('Downloading latest Dusty version: {}'.format(version))
+    if version == constants.VERSION():
+        log_to_client('You\'re already running the latest Dusty version ({})'.format(version))
+        return
+    else:
+        log_to_client('Downloading Dusty version {}'.format(version))
     _download_binary(version)
     _test_dusty_binary(version)
     _move_temp_binary_to_path()
--- a/dusty/commands/upgrade.py
+++ b/dusty/commands/upgrade.py
@@ -1,0 +1,41 @@
+import urllib
+import os
+from subprocess import check_output, check_call
+
+from .. import constants
+from ..daemon import close_connection
+from ..log import log_to_client
+from ..payload import daemon_command
+
+
+def _get_latest_version():
+    """Gets latest Dusty binary version using the GitHub api"""
+    pass
+
+def _get_binary_url(version):
+    return 'https://github.com/gamechanger/dusty/releases/download/{}/dusty'.format(version)
+
+def _get_binary_location():
+    return check_output(['which', 'dusty']).rstrip()
+
+def _download_and_replace_binary(version):
+    binary_url = _get_binary_url(version)
+    binary_location = _get_binary_location()
+    conn = urllib.urlopen(binary_url)
+    if conn.getcode() >= 300:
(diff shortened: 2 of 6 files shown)


dusty/commands/upgrade.py
```diff
@@ -0,0 +1,70 @@
...
+
+from .. import constants
+from ..daemon import close_client_connection
+from ..log import log_to_client
+from ..payload import daemon_command
+
+
+def _get_latest_version():
+    """Gets latest Dusty binary version using the GitHub api"""
+    url = 'https://api.github.com/repos/gamechanger/dusty/releases/latest'
```
thieman: `gamechanger/dusty` should be a constant

dusty/commands/upgrade.py
```diff
@@ -0,0 +1,70 @@
...
+    conn = urllib.urlopen(url)
+    json_data = conn.read()
+    return json.loads(json_data)['tag_name']
```
thieman: This function seems pretty brittle, you're not checking the HTTP status code or anything

dusty/cli/upgrade.py
```diff
@@ -0,0 +1,25 @@
...
+binary with the specified version from GitHub.  If no version is
+specified, the latest version of the binary will be used.  The
+daemon will then exec into the new version of the binary.
+
+Usage:
+  upgrade [<version>]
+
+Options:
+  <version>         If provided, this version of Dusty will be downloaded
+                  and used (defaults to use the most recent version)
```
thieman: text alignment

dusty/cli/upgrade.py
```diff
@@ -0,0 +1,25 @@
+"""Upgrade Dusty's binaries
+
+If Dusty is being run with a binary, this command will replace the
+binary with the specified version from GitHub.  If no version is
+specified, the latest version of the binary will be used.  The
+daemon will then exec into the new version of the binary.
```
thieman: Use more user-facing language here, talk about upgrading the installed version of Dusty, not "the binary"

dusty/commands/upgrade.py
```diff
@@ -0,0 +1,70 @@
...
+    return check_output(['which', 'dusty']).rstrip()
+
+def _download_binary(version):
+    binary_url = _get_binary_url(version)
+    conn = urllib.urlopen(binary_url)
+    if conn.getcode() >= 300:
+        raise RuntimeError('Unable to retrieve Dusty binary version {} from GitHub; this version may not exist'.format(version))
+    binary_data = conn.read()
+    with open(constants.TEMP_BIN_PATH, 'w') as f:
+        f.write(binary_data)
```
thieman: Just use the `tempfile` module for this, gets around permissions errors and assumptions

dusty/commands/upgrade.py
```diff
@@ -0,0 +1,70 @@
...
+    owner = st.st_uid
+    group = st.st_gid
+    shutil.move(constants.TEMP_BIN_PATH, binary_path)
+    os.chown(binary_path, owner, group)
+    os.chmod(binary_path, permissions)
+
+@daemon_command
+def upgrade_dusty_binary(version=None):
+    if not constants.BINARY:
+        log_to_client("The dusty upgrade command only works when dusty is run with its binaries")
```
thieman: More user-facing error message here. Something like "It looks like you're running Dusty from source. Upgrade is only available when running from installed binaries."

dusty/commands/upgrade.py
```diff
@@ -0,0 +1,70 @@
...
+        return
+    if version is None:
+        version = _get_latest_version()
+        log_to_client('Downloading latest Dusty version: {}'.format(version))
```
thieman: Should abort if latest version is equal to what is already installed

dusty/commands/upgrade.py
```diff
@@ -0,0 +1,70 @@
...
+    _download_binary(version)
+    _test_dusty_binary(version)
+    _move_temp_binary_to_path()
+    log_to_client('Finished upgrade to version {} of Dusty!  The daemon will now restart'.format(version))
+    close_client_connection()
+    os.execvp('dusty', ['dusty', '-d'])
```
thieman: I believe you need to use `execvpe` here and pass along the current environment of the daemon process to make sure  you're running with the same config as before.

dusty/commands/upgrade.py
```diff
@@ -0,0 +1,70 @@
...
+    url = 'https://api.github.com/repos/gamechanger/dusty/releases/latest'
+    conn = urllib.urlopen(url)
+    json_data = conn.read()
+    return json.loads(json_data)['tag_name']
+
+def _get_binary_url(version):
+    return 'https://github.com/gamechanger/dusty/releases/download/{}/dusty'.format(version)
+
+def _get_binary_location():
+    return check_output(['which', 'dusty']).rstrip()
```
thieman: I think you can use the proc module we have elsewhere to figure out the binary location of the current process. This which check is insufficient and could fail based on PATH settings

dusty/commands/upgrade.py
```diff
@@ -0,0 +1,70 @@
...
+
+def _get_latest_version():
+    """Gets latest Dusty binary version using the GitHub api"""
+    url = 'https://api.github.com/repos/gamechanger/dusty/releases/latest'
+    conn = urllib.urlopen(url)
+    json_data = conn.read()
+    return json.loads(json_data)['tag_name']
+
+def _get_binary_url(version):
+    return 'https://github.com/gamechanger/dusty/releases/download/{}/dusty'.format(version)
```
thieman: The final filename of the binary should be a constant

dusty/commands/upgrade.py
thieman: Binary filename constant should also be used here

dusty/commands/upgrade.py
jsingle: I think you use the `e` functions to allow you to manually specify the environment, as opposed to just passing the current environment.  That's my reading of the documentation anyway

thieman: You're right

dusty/commands/upgrade.py
```diff
@@ -0,0 +1,76 @@
...
+    os.chmod(binary_path, permissions)
+
+@daemon_command
+def upgrade_dusty_binary(version=None):
+    if not constants.BINARY:
+        log_to_client('It looks like you\'re running Dusty from source. Upgrade is only available when you use the installed binary')
+        return
+    if version is None:
+        version = _get_latest_version()
+    if version == constants.VERSION():
```
thieman: `()`?

jsingle: LOL

dusty/daemon.py
```diff
@@ -46,7 +48,15 @@ def _run_pre_command_functions(connection, suppress_warnings, client_version):
     check_and_load_ssh_auth()
     _refresh_warnings()
 
+def close_client_connection(terminator=SOCKET_TERMINATOR):
+    try:
```
thieman: Need a docstring here to say why this exists

dusty/commands/upgrade.py
```diff
@@ -0,0 +1,84 @@
...
+def _test_dusty_binary(binary_path, version):
+    try:
+        output = check_output([binary_path, '-v']).rstrip()
+    except CalledProcessError:
+        raise RuntimeError('Downloaded binary is not operating correctly; aborting upgrade')
+    test_version = output.split()[-1]
+    if test_version != version:
+        raise RuntimeError('Version of downloaded binary {} does not match expected {}'.format(test_version, version))
+
+def _move_temp_binary_to_path(tmp_binary_path):
```
thieman: Should needs to be unit tested

dusty/commands/upgrade.py
```diff
@@ -0,0 +1,84 @@
...
+    if conn.getcode() >= 300:
+        raise RuntimeError('Unable to retrieve Dusty binary version {} from GitHub; this version may not exist'.format(version))
+    binary_data = conn.read()
+    tmp_path = tempfile.mktemp()
+    with open(tmp_path, 'w') as f:
+        f.write(binary_data)
+    os.chmod(tmp_path, 0755)
+    return tmp_path
+
+def _test_dusty_binary(binary_path, version):
```
thieman: Should be unit-tested against the output you expect to get back from `dusty -v`

jsingle: Unit testing doesn't work super well here because you need a binary path to test

thieman: I'm saying mock out the check_output call to return what you think it should return

jsingle: To me that seems like a very trivial test

thieman: The entire logic is output.split()[-1] assuming the version string is formatted in a particular way.

You're right though, this would make more sense as an integration test.

dusty/commands/upgrade.py
jsingle: Tested a couple things, but testing `chown` in unittests seems hard

PR #347, opened 2015-07-06
add a new faq question and answer

paetling:
review @jsingle

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -33,7 +33,7 @@ ssh-add -K <path-of-private-key>
 This will securely save the passphrase for that key in your Keychain.
 
 ### My tests are hanging at this step: `Creating test...`.  What do I do?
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -32,6 +32,9 @@ ssh-add -K <path-of-private-key>
 ```
 This will securely save the passphrase for that key in your Keychain.
 
+### My tests are hanging at this step: `Creating test...`.  What do I do?
+Dusty tests use docker logs (specifically docker-py's client.logs) functionality.  Docker logs only prints out one line at a time.  So if you are using a test library which outputs its results on a single line (like python's nosetests) this can give the illusion that your tests are hanging when they running fine.
```
jsingle: Mentioning docker-py is probably more low level than we want to mention here.  Maybe 
"Dusty tests use `docker logs`, which stream one line at a time."

paetling: cool

PR #352, opened 2015-07-06
Run rsync as root on remote so we can access all files

thieman:
@jsingle cc @paetling


jsingle:
OK if you like this, please remove the functions that change permissions in systems.virtualbox

PR #359, opened 2015-07-07
Zm integration prep

zacharym:
@paetling this should help when writing integration tests for commands that need to do exec_docker calls cc @thieman


tests/testcases.py
```diff
@@ -20,6 +21,10 @@
 from dusty.path import parent_dir
 from .fixtures import basic_specs_fixture
 
+def exec_docker_patch(*args):
+    args = ['docker'] + [a for a in args]
+    subprocess.check_output(args=args, shell=True)
```
paetling: don't we want to expose or return the output in someway?

zacharym: I am about to do that in a scripts test. I can add it here.

tests/testcases.py
```diff
@@ -118,24 +124,28 @@ def stdout(self):
         clear_stdout was called."""
         return sys.stdout.getvalue()[self.stdout_start:].strip()
 
-    @patch('dusty.commands.utils.exec_docker')
+    def exec_docker_patch(self, *args):
+        args = ['/usr/local/bin/docker'] + [a for a in args]
```
paetling: why can you not just use docker?

tests/testcases.py
```diff
@@ -118,24 +124,28 @@ def stdout(self):
...
-        client_entrypoint()
-        for call in fake_exit.mock_calls:
-            name, args, kwargs = call
-            if len(args) == 1 and args[0] > 0:
-                self._clear_stdout()
-                raise CommandError('Command {} returned with exit code: {}'.format(' '.join(sys.argv), args[0]))
-        result = self.stdout
-        self._clear_stdout()
-        return result
+        with patch('dusty.commands.utils.exec_docker', wraps=self.exec_docker_patch) as fake_exec_docker:
```
paetling: why is this better then @patch on the top?

tests/testcases.py
```diff
@@ -118,24 +124,28 @@ def stdout(self):
...
+            self.fake_exec_docker = fake_exec_docker
```
paetling: why are you setting this?

paetling:
probably worth making some helper functions to get at the logs for the subprocess

tests/testcases.py
zacharym: it doesn't know what self is using the decorator

tests/testcases.py
zacharym: see the helper function `assertExecDocker`, it get's used in the logs tests that are arguably unit tests

tests/testcases.py
zacharym: the subprocess can only user the `docker` alias if you set shell=True, but imo that is a security risk and adds some other complication as well

thieman: you're assuming a binary location here which is basically never okay. use `which` to figure it out.

tests/testcases.py
paetling: ahh

tests/testcases.py
thieman: Those do all sorts of daemon stuff, they aren't unit tests

PR #360, opened 2015-07-07
Js restart revamp

jsingle:
@paetling objective is to allow restarting dead containers.  seems to work

### DIFF
--- a/dusty/systems/docker/__init__.py
+++ b/dusty/systems/docker/__init__.py
@@ -65,10 +65,8 @@ def _get_dusty_containers(client, services, include_exited=False):
     of services. If no services are provided, attempts to
     return all containers associated with Dusty."""
     if services:
-        return [container
-                for container in client.containers(all=include_exited)
-                if any('/{}'.format(get_dusty_container_name(service)) in container.get('Names', [])
-                       for service in services)]
+        containers = [_get_container_for_app_or_service(client, service, include_exited=include_exited) for service in services]
+        return [container for container in containers if container]
     else:
         return [container
                 for container in client.containers(all=include_exited)
@@ -80,6 +78,7 @@ def _get_container_for_app_or_service(client, app_or_service_name, raise_if_not_
             return container
     if raise_if_not_found:
         raise RuntimeError('No running container found for {}'.format(app_or_service_name))
+    return None
 
 def _get_canonical_container_name(container):
     """Return the canonical container name, which should be
@@ -88,6 +87,9 @@ def _get_canonical_container_name(container):
     to which they are linked, but simply taking the shortest name
     should be sufficient to get us the shortest one."""
     return sorted(container['Names'], key=lambda name: len(name))[0][1:]
+
+def _get_app_or_service_name_from_container(container):
+    return _get_canonical_container_name(container).split('_')[1]
 
 def get_dusty_containers(app_or_service_names, include_exited=False):
     client = get_docker_client()
--- a/dusty/systems/docker/compose.py
+++ b/dusty/systems/docker/compose.py
@@ -50,15 +50,14 @@ def _compose_rm(compose_file_location, project_name, services):
         command += services
     check_and_log_output_and_error_demoted(command, env=get_docker_env())
 
-def _check_linked_container_status(client, container, assembled_specs):
+def _check_stopped_linked_containers(client, container, assembled_specs):
+    stopped_containers = []
     app_or_service_name = _get_app_or_service_name_from_container(container)
     linked_containers = links_for_app_or_service(app_or_service_name, assembled_specs)
     for linked_name in linked_containers:
         if _get_dusty_containers(client, [linked_name]) == []:
-            log_to_client('No running container for {0}, which is linked to by {1}.  Cannot restart {1}'.format(
-                              linked_name, app_or_service_name))
-            return False
-    return True
+            stopped_containers.append(linked_name)
+    return stopped_containers
 
 def _compose_restart(services):
     """Well, this is annoying. Compose 1.2 shipped with the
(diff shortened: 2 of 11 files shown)


dusty/systems/docker/__init__.py
```diff
@@ -65,10 +65,8 @@ def _get_dusty_containers(client, services, include_exited=False):
     of services. If no services are provided, attempts to
     return all containers associated with Dusty."""
     if services:
-        return [container
-                for container in client.containers(all=include_exited)
-                if any('/{}'.format(get_dusty_container_name(service)) in container.get('Names', [])
-                       for service in services)]
+        containers = [_get_container_for_app_or_service(client, service, include_exited=include_exited) for service in services]
+        return [container for container in containers if container]
```
paetling: these are functionally equivalent right?

jsingle: The difference is that there's better handling in the case where there's no such container.  So it will print nice stuff to the user if they typo instead of silently just skipping the container that's not found

dusty/systems/docker/compose.py
```diff
@@ -63,14 +63,9 @@ def _restart_container(client, container):
...
-    expected_number_of_containers = get_expected_number_of_running_containers() if len(services) == 0 else len(services)
-    if len(dusty_containers) != expected_number_of_containers:
-        log_to_client("Not going to restart containers. Expected number of containers {} does not match {}".format(expected_number_of_containers, len(dusty_containers)))
-        raise RuntimeError("Please use `docker ps -a` to view crashed containers")
-    else:
-        for container in dusty_containers:
-            _restart_container(client, container)
+    dusty_containers = _get_dusty_containers(client, services, include_exited=True)
+    for container in dusty_containers:
+        _restart_container(client, container)
```
paetling: so all you are really doing here is taking out the check case for if containers have died.   That was put in for a reason. Containers depend on containers they link to being started before they attempt to restart.  This no longer accounts for that case.

dusty/systems/docker/__init__.py
paetling: cool

dusty/systems/docker/compose.py
jsingle: You're right.  Is that the only reason the check was there?  Because I can implement a checker to see if containers to link to are started

paetling: yes.  And i like that idea.

jsingle: Given this issue, why do we allow people to restart all their containers?  Won't that hit this same issue?

jsingle: Well it doesn't seem to be an issue for restarting containers

dusty/systems/docker/compose.py
```diff
@@ -3,11 +3,14 @@
 
 import yaml
 
-from . import _get_canonical_container_name, get_docker_env, get_docker_client, _get_dusty_containers
+from . import (_get_canonical_container_name, get_docker_env, get_docker_client,
+               _get_dusty_containers, _get_app_or_service_name_from_container,
+               _get_container_for_app_or_service)
```
paetling: do not import _ functions.  Change them to be non underscore.

dusty/systems/docker/compose.py
```diff
@@ -47,6 +50,16 @@ def _compose_rm(compose_file_location, project_name, services):
         command += services
     check_and_log_output_and_error_demoted(command, env=get_docker_env())
 
+def _check_linked_container_status(client, container, assembled_specs):
+    app_or_service_name = _get_app_or_service_name_from_container(container)
+    linked_containers = links_for_app_or_service(app_or_service_name, assembled_specs)
+    for linked_name in linked_containers:
+        if _get_dusty_containers(client, [linked_name]) == []:
+            log_to_client('No running container for {0}, which is linked to by {1}.  Cannot restart {1}'.format(
```
paetling: i think this logic about informing the user if its linked container is dead should be in the calling function.  Seems like a check_linked_container_status should just return whether or not linked containers are alive

jsingle: I wanted to log the specific linked container that is not running.  But I guess that can be returned by this function... it just might be a little messy

dusty/systems/docker/compose.py
jsingle: We're already doing that a lot from within the same `systems.docker` module

dusty/systems/docker/compose.py
paetling: You can log the container that is not running and what it is linked to. BUt you should not log the restart here.  seems like a good compromise

dusty/systems/docker/compose.py
paetling: that is not a good excuse for bad coding practices.  Just means we did not catch it in those other places

jsingle: Going through and fixing all these underscores sounds like a task for another PR

paetling: i'm fine with that.

dusty/systems/docker/compose.py
```diff
@@ -61,15 +73,21 @@ def _restart_container(client, container):
...
-        for container in dusty_containers:
+    for service in services:
+        container = _get_container_for_app_or_service(client, service, include_exited=True)
+        if container is None:
+            log_to_client('No container found for {}'.format(service))
+            continue
+        stopped_linked_containers = _check_stopped_linked_containers(client, container, assembled_specs)
+        if stopped_linked_containers:
+            log_to_client('No running containers {0}, which are linked to by {1}.  Cannot restart {1}'.format(
+                              stopped_linked_containers, service))
```
paetling: so what if a container has no links? won't this now be able to restart that container?

paetling:
also might squash the first three commits down to 1

dusty/systems/docker/compose.py
jsingle: yes?  what's wrong with that?

jsingle:
Sure that's reasonable

dusty/systems/docker/compose.py
paetling: i meant not.  not now

PR #361, opened 2015-07-07
Add integration tests for setup CLI command

thieman:
@paetling Also adds a flag for manually setting VM memory size

### DIFF
--- a/tests/integration/cli/setup_test.py
+++ b/tests/integration/cli/setup_test.py
@@ -17,7 +17,7 @@ class TestSetupCLI(DustyIntegrationTestCase):
     def test_setup_defaults(self, fake_raw_input, fake_vm_size):
         fake_vm_size.return_value = 2048
         fake_raw_input.side_effect = ['y', '', 'y']
-        self.run_command('setup')
+        self.run_command('setup --no-update')
         self.assertConfigValue(constants.CONFIG_MAC_USERNAME_KEY, self.current_user)
         self.assertConfigValue(constants.CONFIG_SPECS_REPO_KEY, 'github.com/gamechanger/dusty-example-specs')
         self.assertConfigValue(constants.CONFIG_VM_MEM_SIZE, 2048)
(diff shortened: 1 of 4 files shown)


tests/integration/cli/setup_test.py
```diff
@@ -0,0 +1,51 @@
...
+
+    @patch('dusty.commands.setup._get_raw_input')
+    def test_setup_override_user(self, fake_raw_input):
+        fake_raw_input.side_effect = ['n', self.current_user, '', 'y']
+        self.run_command('setup')
+        self.assertConfigValue(constants.CONFIG_MAC_USERNAME_KEY, self.current_user)
+
+    @patch('dusty.commands.setup._get_raw_input')
+    def test_setup_override_specs(self, fake_raw_input):
+        fake_raw_input.side_effect = ['y', 'github.com/gamechanger/dusty-example-specs', 'y']
```
paetling: i think it is a more thorough test to check that both this one and the test above set the values to something other then self.current_user and dusty-example-specs.  That means that across the first and the second two tests we are ensuring that we are not just always writing those values to the config in setup.

tests/integration/cli/setup_test.py
```diff
@@ -0,0 +1,51 @@
...
+        fake_raw_input.side_effect = ['y', '', 'n', '1024']
+        self.run_command('setup')
+        self.assertConfigValue(constants.CONFIG_VM_MEM_SIZE, 1024)
+
+    def test_setup_flags(self):
+        self.run_command('setup --mac_username={} --default_specs_repo=github.com/gamechanger/dusty-example-specs --nginx_includes_dir=/tmp/nginx --boot2docker_vm_memory=1024'.format(self.current_user))
+        self.assertConfigValue(constants.CONFIG_MAC_USERNAME_KEY, self.current_user)
+        self.assertConfigValue(constants.CONFIG_SPECS_REPO_KEY, 'github.com/gamechanger/dusty-example-specs')
+        self.assertConfigValue(constants.CONFIG_VM_MEM_SIZE, 1024)
+        self.assertConfigValue(constants.CONFIG_NGINX_DIR_KEY, '/tmp/nginx')
```
paetling: similar thoughts here

paetling:
:ghost: :cow:

PR #374, opened 2015-07-08
Js fix for app w no repo

jsingle:
@paetling we don't require apps to actually have repos, but this code path assumed that they do


paetling:
:ghost: :cow:

PR #376, opened 2015-07-08
Js app service lib name unique

jsingle:
@thieman add validation to ensure that apps do not share the same name as services or libs

### DIFF
--- a/dusty/commands/validate.py
+++ b/dusty/commands/validate.py
@@ -47,6 +47,9 @@ def _check_name_overlap(specs):
     app_lib_overlap = apps.intersection(libs)
     if app_lib_overlap:
         raise ValidationException('Apps and libs cannot share names: {}'.format(app_lib_overlap))
+    service_lib_overlap = services.intersection(libs)
+    if service_lib_overlap:
+        raise ValidationException('Services and libs cannot share names: {}'.format(service_lib_overlap))
 
 def _validate_spec_names(specs):
     _check_name_overlap(specs)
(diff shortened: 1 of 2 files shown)


dusty/commands/validate.py
```diff
@@ -37,7 +37,19 @@ def _validate_lib_references(lib, specs):
...
+def _check_name_overlap(specs):
+    apps = set(specs['apps'].keys())
+    libs = set(specs['libs'].keys())
+    services = set(specs['services'].keys())
+    app_service_overlap = apps.intersection(services)
+    if app_service_overlap:
+        raise ValidationException('Apps and services cannot share names: {}'.format(app_service_overlap))
+    app_lib_overlap = apps.intersection(libs)
+    if app_lib_overlap:
+        raise ValidationException('Apps and libs cannot share names: {}'.format(apps_lib_overlap))
```
thieman: Don't we just want to make all names unique across all types of specs?

jsingle: Well I'd say it makes sense to permit bundles to share names with apps.  We don't strictly need to check anything more than this to ensure functionality

thieman: Services and libs sharing names are ok? Seems weird.

jsingle:
Services and libs seems weird but I didn't think it was worth explicitly checking it and raising an exception.  Don't mind fixing that though

thieman:
Figure we might as well check it.

PR #377, opened 2015-07-08
Check for git executable in preflight

### LINKED ISSUE #368: Add git to preflight checks
The GitPython library we're using requires access to a `git` binary under the hood. We should add this as an explicit check we do during preflight to ensure the daemon can find and access the binary.

thieman:
@jsingle Also logs the PATH, which should be useful for debugging when the preflight fails.

Closes #368.


jsingle:
Seems reasonable :zap:

PR #378, opened 2015-07-08
More clarity around SSH key failures

### LINKED ISSUE #367: ssh-add error message is incorrect
Currently, the error message around Git failures suggests that you should only see it when trying to access private repositories:

```
ERROR: Git command failed. If you are trying to access a private repo, please make sure you have added your SSH key to the SSH agent using: ssh-add <SSH key filepath>
```

This is inaccurate. Since we're doing all remote Git access over SSH, you need this even if you're trying to get to a public repository (like the example specs).

thieman:
@paetling Closes #367.

### DIFF
--- a/dusty/source.py
+++ b/dusty/source.py
@@ -17,9 +17,9 @@ def git_error_handling():
     try:
         yield
     except git.exc.GitCommandError:
-        log_to_client('ERROR: Git command failed. If you are trying to access a private repo, '
-                      'please make sure you have added your SSH key to the SSH agent using: '
-                      'ssh-add <SSH key filepath>')
+        log_to_client('ERROR: Git command failed. If you are trying to access a remote repository '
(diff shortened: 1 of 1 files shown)


dusty/source.py
```diff
@@ -17,9 +17,9 @@ def git_error_handling():
     try:
         yield
     except git.exc.GitCommandError:
-        log_to_client('ERROR: Git command failed. If you are trying to access a private repo, '
-                      'please make sure you have added your SSH key to the SSH agent using: '
-                      'ssh-add <SSH key filepath>')
+        log_to_client('ERROR: Git command failed. If you are trying to access a remote repository '
+                      'over SSH (e.g. GitHub), please make sure you have added your SSH key to the '
```
paetling: do we need the SSH line.  If it is remote, don't we use ssh in dusty

thieman: We do for now, aren't we planning on adding HTTPS soon?

PR #379, opened 2015-07-08
add test integration tests

paetling:
review @thieman I added the a handler which can be used in tests to get the output from client calls to log_to_client.  Using this I wrote integration tests testing the test cli command

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -93,16 +93,26 @@ def specs_fixture_with_depends():
     _write('service', 'servica', {'image': 'busybox'})
 
 
-def busybox_single_app_bundle_fixture(num_bundles=1, command=['sleep 999999999']):
+def busybox_single_app_bundle_fixture(num_bundles=1, command=['sleep 999999999'], include_tests=False):
     """Fixture for use in integration tests. The local repo at
     /tmp/fake-repo should be set up before using this fixture."""
+    app_dict = {'repo': '/tmp/fake-repo',
+                'mount': '/repo',
+                'image': 'busybox',
+                'commands': {'always': command}}
+    if include_tests:
+        app_dict['test'] = {'image': 'python',
+                            'once': ['pip install nose'],
+                            'suites': [{'name': 'test1',
+                                        'command': ['nosetests'],
+                                        'default_args': '.'},
+                                        {'name': 'test2',
+                                        'command': ['nosetests'],
+                                        'default_args': '.'}]}
     for bundle in range(num_bundles):
         name = 'busybox{}'.format(_num_to_alpha(bundle))
         _write('bundle', name, {'description': 'Busybox bundle', 'apps': [name]})
-        _write('app', name, {'repo': '/tmp/fake-repo',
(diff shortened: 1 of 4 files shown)


tests/fixtures.py
```diff
@@ -93,16 +93,26 @@ def specs_fixture_with_depends():
...
 
-def busybox_single_app_bundle_fixture(num_bundles=1, command=['sleep 999999999']):
+def busybox_single_app_bundle_fixture(num_bundles=1, command=['sleep 999999999'], include_tests=False):
     """Fixture for use in integration tests. The local repo at
     /tmp/fake-repo should be set up before using this fixture."""
+    app_dict = {'repo': '/tmp/fake-repo',
+                'mount': '/repo',
+                'image': 'busybox',
+                'commands': {'always': command}}
+    if include_tests:
```
thieman: Why does this need to be optional?

tests/integration/cli/test_test.py
```diff
@@ -0,0 +1,46 @@
...
+        self.assertEqual(self.handler.log_to_client_output.count('OK'), 1)
+        self.assertEqual(self.handler.log_to_client_output.count('Running commands to create new image:'), 1)
+
+    def test_basic_test_args(self):
+        try:
+            self.run_command('test --recreate busyboxa test1 test/location')
+        except:
+            pass
+        self.assertEqual(self.handler.log_to_client_output.count('TESTS FAILED'), 1)
+        self.assertEqual(self.handler.log_to_client_output.count('OSError: No such file'), 1)
```
thieman: This is kinda jank, is there a more specific / less brittle way to test that your args did something?

tests/integration/cli/test_test.py
```diff
@@ -0,0 +1,46 @@
...
+from ...testcases import DustyIntegrationTestCase
+from ...fixtures import busybox_single_app_bundle_fixture
+from dusty.log import client_logger, DustyClientTestingSocketHandler
+
+class TestTestCLI(DustyIntegrationTestCase):
+    def setUp(self):
+        super(TestTestCLI, self).setUp()
+        busybox_single_app_bundle_fixture(num_bundles=1, include_tests=True)
+        self.handler = DustyClientTestingSocketHandler()
+        client_logger.addHandler(self.handler)
```
thieman: Should this be in the general setUp for DustyIntegrationTestCase?

tests/fixtures.py
paetling: it probably does not, just did not want to step on other tests toes.

tests/integration/cli/test_test.py
paetling: yes.  my test2 could be something like `ls` and I could list a different location.

tests/integration/cli/test_test.py
paetling: interesting question. It does not technically need to be, it probably should be just to help people in the future.

tests/fixtures.py
thieman: Would be simpler to not make it optional.

Issue #380, opened 2015-07-08
Backgrounded tasks in once script block execution until they finish


thieman:
I'm trying to do something like this in an app spec:

```yaml
commands:
  once:
    - (apt-get update && apt-get install -y nodejs npm) &
    - ...
```

My expectation is that this line would work in the background and that the execution of the `once` command would not wait for it to complete. Instead, my container waits for this backgrounded task to complete before moving on to the `always` command during container startup.

Ping @jsingle who claims to know what's wrong here

jsingle:
Solved by https://github.com/gamechanger/dusty/pull/385

PR #385, opened 2015-07-09
Js cleanup named pipes

jsingle:
@paetling PR does:
 * Remove `wait` on pid, which was causing to block on background threads. 
 * Don't send output of `always` to a log file.  I don't think we need to do this since that's the main command that gets run on the container (just use dusty logs)

### DIFF
--- a/dusty/command_file.py
+++ b/dusty/command_file.py
@@ -52,7 +52,7 @@ def _get_always_commands(app_spec):
         commands_with_function.append('dusty_always_fn () {')
         commands_with_function += always_commands
         commands_with_function.append('}')
-        commands_with_function += _tee_output_commands('dusty_always_fn')
+        commands_with_function.append('dusty_always_fn')
     return commands_with_function
 
 def _compile_docker_commands(app_name, assembled_specs):
--- a/dusty/command_file.py
+++ b/dusty/command_file.py
@@ -23,7 +23,7 @@ def _tee_output_commands(command_to_tee):
         'rm -f $PIPEFILE',
         'mkfifo $PIPEFILE',
         '(tee {}.log < $PIPEFILE || true; rm -f $PIPEFILE) &'.format(os.path.join(constants.CONTAINER_LOG_PATH, command_to_tee)),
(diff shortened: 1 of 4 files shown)


dusty/command_file.py
```diff
@@ -22,11 +22,8 @@ def _tee_output_commands(command_to_tee):
         'PIPEFILE={}_pipe_file'.format(tee_function_name),
         'rm -f $PIPEFILE',
         'mkfifo $PIPEFILE',
-        'tee {}.log < $PIPEFILE &'.format(os.path.join(constants.CONTAINER_LOG_PATH, command_to_tee)),
-        'TEEPID=$!',
+        '(tee {}.log < $PIPEFILE || true; rm -f $PIPEFILE) &'.format(os.path.join(constants.CONTAINER_LOG_PATH, command_to_tee)),
```
paetling: why do you need the || true?  wont the result of this statement just be the command after the ;?

paetling:
I do not really understand your argument for the second part. Can't you also look at logs for once?  I guess since it does not always run it is not worth saving?

dusty/command_file.py
jsingle: nope because we `set -e`, which anecdotally persists to this sub shell thingy

jsingle:
My argument is that `once` output is worth saving.  `always` output is not

dusty/command_file.py
paetling: But if outputting to the log fails, don't we want the whole thing to fail?

jsingle: It won't anyway because it's run in the background.  And that'd debateable

paetling:
I like the once always distinction. I think we mention in our logs somewhere about them. Worth a quick glance and correction if it is there.

dusty/command_file.py
paetling: i see.   ok.

paetling:
just the logs thing and then :ghost: :cow:

jsingle:
Closes https://github.com/gamechanger/dusty/issues/380

PR #386, opened 2015-07-09
add in integration tests for up

paetling:
review @thieman

### DIFF
--- a/tests/integration/cli/up_test.py
+++ b/tests/integration/cli/up_test.py
@@ -1,0 +1,34 @@
+import subprocess
+
+from ...testcases import DustyIntegrationTestCase
+from ...fixtures import busybox_single_app_bundle_fixture
+from dusty.systems.docker import get_docker_env
+
+class TestUpCLI(DustyIntegrationTestCase):
+    def setUp(self):
+        super(TestUpCLI, self).setUp()
+        busybox_single_app_bundle_fixture(num_bundles=2)
+        self.run_command('bundles activate busyboxa busyboxb')
+
+    def test_basic_up_command(self):
+        run_output = self.run_command('up')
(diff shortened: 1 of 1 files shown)


tests/integration/cli/up_test.py
```diff
@@ -0,0 +1,34 @@
...
+
+class TestUpCLI(DustyIntegrationTestCase):
+    def setUp(self):
+        super(TestUpCLI, self).setUp()
+        busybox_single_app_bundle_fixture(num_bundles=2)
+        self.run_command('bundles activate busyboxa busyboxb')
+
+    def test_basic_up_command(self):
+        run_output = self.run_command('up')
+        self.assertTrue('Your local environment is now started!' in run_output)
```
thieman: Use `self.assertIn`

tests/integration/cli/up_test.py
```diff
@@ -0,0 +1,34 @@
...
+
+        docker_output = subprocess.check_output(['sh', '-c', "docker ps"], env=get_docker_env())
```
thieman: You generally want to avoid shelling out unless you absolutely need to. You could do this with the Docker driver directly, which also has the advantage of being less brittle. The driver will get raw JSON back and interpret it. Shelling out relies on the command line output of a function which maintainers generally think can change without making a big deal about it.

paetling: This is a good point

tests/integration/cli/up_test.py
paetling: will change

tests/integration/cli/up_test.py
thieman: Also this seems like it should be a general assert function provided by the test case. Could probably even piggyback on the existing inspect method to do it easily.

PR #387, opened 2015-07-09
Enable integration tests to run with non-root mac_username

thieman:
@paetling @jsingle @zacharym Can now provide `DUSTY_INTEGRATION_TESTS_USER` environment var to run with your normal mac_username. Should hopefully eliminate the pain around running integration tests locally. Only necessary change was some permission fudging.

### DIFF
--- a/tests/integration/cli/repos_test.py
+++ b/tests/integration/cli/repos_test.py
@@ -35,9 +35,9 @@ class TestReposCLI(DustyIntegrationTestCase):
     def tearDown(self):
         self.run_command('bundles deactivate busyboxa')
         constants.REPOS_DIR = self.old_repos_dir
-        rmtree(self.fake_override_repo_location)
-        rmtree(self.fake_from_dir)
         rmtree(self.temp_repos_dir)
+        rmtree(self.fake_override_dir)
+        rmtree(self.fake_from_dir)
         super(TestReposCLI, self).tearDown()
 
     def test_repos_list(self):
--- a/tests/integration/cli/repos_test.py
+++ b/tests/integration/cli/repos_test.py
@@ -1,4 +1,4 @@
-from os import path
+import os
 from shutil import rmtree
 from subprocess import check_call
 from tempfile import mkdtemp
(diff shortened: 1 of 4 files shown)


thieman:
Once this gets merged would recommend blowing away any root VM you have and running this way from now on

tests/integration/cli/repos_test.py
```diff
@@ -16,13 +16,20 @@ def setUp(self):
...
         busybox_single_app_bundle_fixture(num_bundles=1)
         self.run_command('bundles activate busyboxa')
+
         self.temp_repos_dir = mkdtemp()
+        self.fake_override_dir = mkdtemp()
+        self.fake_from_dir = mkdtemp()
+
+        os.chmod(self.temp_repos_dir, 0777)
+        os.chmod(self.fake_override_dir, 0777)
+        os.chmod(self.fake_from_dir, 0777)
```
paetling: why do you need to change these permissions?

tests/testcases.py
```diff
@@ -89,7 +89,10 @@ def setUp(self):
         write_default_config()
         save_config_value(constants.CONFIG_SETUP_KEY, True)
         save_config_value(constants.CONFIG_SPECS_REPO_KEY, 'github.com/gamechanger/dusty-example-specs')
-        save_config_value(constants.CONFIG_MAC_USERNAME_KEY, self.current_user)
+
+        self.tests_user = os.getenv('DUSTY_INTEGRATION_TESTS_USER', self.current_user)
```
paetling: won't this get the value for the environment variable and use that, if it cannot find it then it uses self.current_user?

jsingle: Yeah that should be what we want right?

tests/integration/cli/repos_test.py
zacharym: made me notice, should tearDown instead `rmtree(fake_override_dir)` rather than `fake_override_repo_location` which lives inside `fake_override_dir`? Alternatively, would it work to just `chmod(fake_override_repo_location)`?

tests/testcases.py
paetling: I see.  I misread what the comment above meant. I thought it meant that if you provide taht environment variable it will use your local user.  What it means is that if you set that variable to your local user, life is good.

jsingle:
This PR makes me happy

paetling:
Also, we directly write to your config in tests.  How is this getting by that?

paetling:
If we think it is better to run it this way, why do we not just default to running it with your current user and then you have to provide the environment variable to change that?

jsingle:
good question

zacharym:
one comment but about the repos test setUp / tearDown but yeah, LGTM

thieman:
@paetling The current user is always root. You need to run the actual test suite itself as root for it to be able to fuck with hosts and stuff.

paetling:
:ghost: :cow:

PR #396, opened 2015-07-10
Add some standard proxy headers to Nginx location blocks

thieman:
@jsingle Found some issues with running iPython, I think adding these in makes Nginx play a little nicer with its upstreams.


jsingle:
Yay ipython.  Seems good to me

PR #397, opened 2015-07-10
add tests for providing args and flags

zacharym:
@jsingle added some more script tests


jsingle:
:shipit:

zacharym:
:itwasn'tme: !

PR #398, opened 2015-07-10
Upgrade integration tests

jsingle:
@thieman review

### DIFF
--- a/tests/upgrade_integration/upgrade_test.py
+++ b/tests/upgrade_integration/upgrade_test.py
@@ -44,17 +44,18 @@ class TestUpgrade(DustyIntegrationTestCase):
 
     def test_upgrade_source_fails(self):
         self.run_daemon_source()
-        output = ''
         output = self.run_command('upgrade')
         self.assertTrue('It looks like you\'re running Dusty from source' in output)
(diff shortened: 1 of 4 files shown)


tests/upgrade_integration/upgrade_test.py
```diff
@@ -0,0 +1,61 @@
...
+    def test_upgrade_2_1(self):
+        self.run_daemon_binary()
+        version = '0.2.1'
+        output = self.run_command('version')
+        self.assertInSameLine(output, 'daemon', 'version')
+        output = self.run_command('upgrade {}'.format(version))
+        self.assertInSameLine(output, 'Downloading', version)
+        self.assertInSameLine(output, 'Finished upgrade', version)
+        sleep(1)
+        output = self.run_command('version', raise_on_error=False)
```
thieman: Can't you just catch this instead of adding this `raise_on_error` flag you only use for testing?

jsingle: If I do that I can't see the output of the command

thieman: ok

Issue #366, opened 2015-07-08
Having to ssh-add is confusing


thieman:
Our limited experience with public release has shown that most people don't already have their SSH keys added to their `ssh-agent`. This makes it hard to get started, and pretty much guarantees the first thing they see out of Dusty is a stacktrace.

We should make this an explicit part of setup and try to guide people towards the right solution.

paetling:
closed by https://github.com/gamechanger/dusty/pull/409

PR #404, opened 2015-07-13
Ae check ssh config

paetling:
review @thieman we decided the best we to address #366 is to more explicitly tell the user they need to add on setup.


dusty/commands/setup.py
```diff
@@ -84,6 +85,13 @@ def _get_boot2docker_vm_size():
     else:
         return _get_raw_input('Please input the number of megabytes to allocate to the vm: ')
 
+def _check_ssh_add():
+    output = ''
+    try:
+        check_call_demoted(['sh', '-c', 'ssh-add -l'])
```
thieman: Why do we need a shell here? Also don't we always do Git commands as root? If so, this shouldn't be demoted.

dusty/commands/setup.py
```diff
@@ -109,6 +117,8 @@ def setup_dusty_config(mac_username=None, specs_repo=None, nginx_includes_dir=No
     else:
         boot2docker_vm_memory = _get_boot2docker_vm_size()
 
+    _check_ssh_add()
```
thieman: If we're telling the user to do things, it should be the last thing we output from the setup command, should be very visible (####HEY LOOK AT ME####), and should have explicit instructions on what to do.

dusty/commands/setup.py
```diff
@@ -84,6 +85,13 @@ def _get_boot2docker_vm_size():
     else:
         return _get_raw_input('Please input the number of megabytes to allocate to the vm: ')
 
+def _check_ssh_add():
+    output = ''
+    try:
+        check_call_demoted(['sh', '-c', 'ssh-add -l'])
+    except:
+        log_to_client('\nYou must run `ssh-add -K` in order to add your ssh keys to your keychain. Dusty relies on this to pull repos. \n')
```
thieman: The wording here makes it seem like I just punch `ssh-add -K` into a terminal. Need to be more explicit that you give it the path to the key.

paetling: You do not need to give the path to the key generally.  If they have a more complicated ssh config setup i am assuming they would know how to do more complicated commands with ssh-add

dusty/commands/setup.py
paetling: The problem with moving it is that we fail to update your git repos without this value being set.  Which means we get this error: 
```
ERROR: Git command failed. If you are trying to access a remote repository over SSH (e.g. GitHub), please make sure you have added your SSH key to the SSH agent using: ssh-add -K <SSH key filepath>
ERROR: 'Error when fetching: fatal: Could not read from remote repository.' returned with exit code 2
```

dusty/commands/setup.py
paetling: Originally we needed to do this demoted because of how add sources the files it adds from. I guess that is not strictly needed here.

dusty/commands/setup.py
thieman: It just defaults to `id_rsa`?

PR #407, opened 2015-07-13
Zm manage   all

zacharym:
@paetling cc @thieman

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## 0.2.2 (In Progress)
- * NEW: `dusty repos manage` can now accept an `--all` flag, to have dusty manage all overriden repos
+ * NEW: `dusty repos manage` can now accept an `--all` flag, to have Dusty manage all overridden repos
  * NEW: `dusty restart` can now restart stopped containers.
  * NEW: `dusty upgrade` command.  This command will upgrade Dusty's version by replacing the Dusty binary
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 0.2.2 (In Progress)
+ * NEW: `dusty repos manage` can now accept an `--all` flag, to have dusty manage all overriden repos
  * NEW: `dusty restart` can now restart stopped containers.
  * NEW: `dusty upgrade` command.  This command will upgrade Dusty's version by replacing the Dusty binary
 
--- a/dusty/cli/repos.py
+++ b/dusty/cli/repos.py
@@ -9,7 +9,7 @@ is useful for actively developing apps and libs that depend on that
 repo. To override a repo, use the `override` or `from` commands.
 
 Usage:
-  repos from [--all] <source_path>
+  repos from <source_path>
   repos list
   repos manage (--all | <repo_name>)
   repos override <repo_name> <source_path>
@@ -18,7 +18,7 @@ Usage:
 Commands:
   from        Override all repos from a given directory
   list        Show state of all repos referenced in specs
-  manage      Tell Dusty to manage a repo, removing any overrides
+  manage      Tell Dusty to manage a repo or all repos, removing any overrides
   override    Override a repo with a local copy that you manage
   update      Pull latest master on Dusty-managed repos
 """
--- a/dusty/cli/repos.py
+++ b/dusty/cli/repos.py
@@ -9,9 +9,9 @@ is useful for actively developing apps and libs that depend on that
 repo. To override a repo, use the `override` or `from` commands.
 
 Usage:
-  repos from <source_path>
+  repos from [--all] <source_path>
   repos list
-  repos manage <repo_name>
+  repos manage (--all | <repo_name>)
   repos override <repo_name> <source_path>
   repos update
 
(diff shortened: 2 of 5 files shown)


dusty/cli/repos.py
```diff
@@ -9,9 +9,9 @@
 repo. To override a repo, use the `override` or `from` commands.
 
 Usage:
-  repos from <source_path>
+  repos from [--all] <source_path>
   repos list
-  repos manage <repo_name>
+  repos manage (--all | <repo_name>)
   repos override <repo_name> <source_path>
   repos update
```
paetling: add an Option:  that describes what --all does.  I do not know from what is above

dusty/cli/repos.py
```diff
@@ -9,9 +9,9 @@
 repo. To override a repo, use the `override` or `from` commands.
 
 Usage:
-  repos from <source_path>
+  repos from [--all] <source_path>
```
jsingle: what does `--all` mean when applied to `from`?  seems like this shouldn't be here

paetling: does not seem like we are using this in the from command?

tests/integration/cli/repos_test.py
```diff
@@ -61,6 +60,21 @@ def test_repos_manage(self):
         self.assertNotInSameLine(result, 'fake-repo', self.fake_override_repo_location)
         self.assertInSameLine(result, 'fake-repo', self.fake_local_repo_location)
 
+    def test_repos_manage_all_with_one(self):
+        self.run_command('repos override fake-repo {}'.format(self.fake_override_repo_location))
+        self.run_command('repos manage --all')
+        result = self.run_command('repos list')
+        self.assertNotInSameLine(result, self.fake_override_repo_location, 'fake-repo')
```
paetling: probably worth asserting they are in the same line above the mange --all command.

tests/integration/cli/repos_test.py
```diff
@@ -61,6 +60,21 @@ def test_repos_manage(self):
...
+
+
+    def test_repos_manage_all_with_multiple(self):
+        self.run_command('repos override fake-repo {}'.format(self.fake_override_repo_location))
+        self.run_command('repos override github.com/app/a {}'.format(self.fake_from_repo_location))
+        self.run_command('repos manage --all')
+        result = self.run_command('repos list')
+        self.assertNotInSameLine(result, self.fake_override_repo_location, 'fake-repo')
+        self.assertNotInSameLine(result, self.fake_from_repo_location, 'github.com/app/a')
```
paetling: same here.

paetling:
is it worth having / adding some tests to repos_test.py?

dusty/cli/repos.py
zacharym: yeah that was just a mistake

tests/unit/commands/repos_test.py
```diff
@@ -65,6 +65,15 @@ def test_override_then_manage_repo(self):
...
                               {get_specs_repo().remote_path: self.temp_specs_path})
 
+    def test_override_then_manage_all(self):
+        override_repo('github.com/app/a', self.temp_specs_path)
+        self.assertItemsEqual(get_config_value(constants.CONFIG_REPO_OVERRIDES_KEY),
+                              {get_specs_repo().remote_path: self.temp_specs_path,
+                               'github.com/app/a': self.temp_specs_path})
+        manage_all_repos()
+        self.assertItemsEqual(get_config_value(constants.CONFIG_REPO_OVERRIDES_KEY),
+                              {})
```
paetling: you are testing the wrong stuff here.  You function (manage_all_repos) just calls manage_repo a bunch of times.  The thing you should be testing is that given some set of repos, mange_repo is called with each of those.   manage_all_repos should not have to know that this call means that the config value will be changed.  That is a different functions responsibly.  In this unit test you are purely testing the execution of this function

dusty/cli/repos.py
```diff
@@ -11,14 +11,14 @@
 Usage:
   repos from <source_path>
   repos list
-  repos manage <repo_name>
+  repos manage (--all | <repo_name>)
   repos override <repo_name> <source_path>
```
paetling: I still think it is a good idea to add this:
```
Options:
--all   When this is specified all repos will be controlled by dusty 
```
or something similar

docs/changelog.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 0.2.2 (In Progress)
+ * NEW: `dusty repos manage` can now accept an `--all` flag, to have dusty manage all overriden repos
```
thieman: typo

dusty/commands/repos.py
```diff
@@ -28,14 +28,23 @@ def override_repo(repo_name, source_path):
...
-def manage_repo(repo_name):
+def _manage_repo(repo_name):
     repo = Repo.resolve(get_all_repos(), repo_name)
     config = get_config_value(constants.CONFIG_REPO_OVERRIDES_KEY)
     if repo.remote_path in config:
         del config[repo.remote_path]
-    save_config_value(constants.CONFIG_REPO_OVERRIDES_KEY, config)
-    log_to_client('Will manage repo {} with Dusty-managed copy of source'.format(repo.remote_path))
+        save_config_value(constants.CONFIG_REPO_OVERRIDES_KEY, config)
+        log_to_client('Will manage repo {} with Dusty-managed copy of source'.format(repo.remote_path))
```
thieman: This will now print absolutely nothing if the user attempts to manage a repo that's already managed. Doesn't seem good for usability.

dusty/commands/repos.py
```diff
@@ -28,14 +28,24 @@ def override_repo(repo_name, source_path):
...
+    log_to_client('No overriden repos found by name {}'.format(repo_name))
```
thieman: Now this gets logged no matter what

Issue #389, opened 2015-07-09
Dusty setup should default to Y generally


thieman:
From @tleach. Currently at least some of the `dusty setup` steps default to no if you hit enter without providing any input. In the case of the first username prompt, at least, that's pretty counterintuitive.

paetling:
close by https://github.com/gamechanger/dusty/pull/408

PR #408, opened 2015-07-13
default setup answers to yes not no

paetling:
review @jsingle this address #389

### DIFF
--- a/tests/unit/commands/setup_test.py
+++ b/tests/unit/commands/setup_test.py
@@ -123,7 +123,7 @@ class TestSetupCommands(DustyTestCase):
     def test_get_and_configure_nginx_includes_dir_usr_no(self, fake_get_contents, fake_isfile, fake_get_raw_input, fake_setup_nginx):
         fake_isfile.side_effect = self.factory_file_side_effect('/usr/local/etc/nginx/nginx.conf')
         fake_get_contents.return_value = ['servers/*;']
-        fake_get_raw_input.return_value = ''
+        fake_get_raw_input.return_value = 'n'
         result = _get_and_configure_nginx_includes_dir()
         self.assertEqual(result, '')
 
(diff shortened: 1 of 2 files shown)


jsingle:
Not sure this is the right way to do it... I think we want a *blank* input to be the same as a Y.  Anything else is invalid (maybe should prompt for `please enter Y or N`).  I feel like that is more standard/intuitive

paetling:
This is a good point. if they input x that should not be interpreted as a yes.   I do not think the messaging needs to change though.  I think it is more just ensuring it is y or ''

jsingle:
OK I don't feel strongly that we need to loop input until Y or N

jsingle:
Also you might want to make a new function which gets raw input and compares for Y or N since you're doing that in 3 places

paetling:
ahead of you on that.  i made a new function to check for if something is a yes input

jsingle:
:cow: :gun: :camel:

paetling:
closes #389

tests/unit/commands/setup_test.py
```diff
@@ -318,3 +318,8 @@ def test_get_boot2docker_vm_size_less_8_n(self, fake_virtual_memory, fake_get_ra
         self.assertEqual(_get_boot2docker_vm_size(), 1)
         fake_get_raw_input.assert_has_calls([call('Your system seems to have 6144 megabytes of memory. We would like to allocate 2048 to your vm. Is that ok? (y/n) '),
                                              call('Please input the number of megabytes to allocate to the vm: ')])
+    @patch('dusty.commands.setup._get_raw_input')
+    def test_enter_is_accepted_as_yes(self, fake_get_raw_input):
+        fake_get_raw_input.return_value = ''
+        setup_dusty_config()
```
thieman: Shouldn't you do asserts at the end of this to confirm it did what you think it should do?

PR #409, opened 2015-07-13
attempt to make docs a bit more clear with respect to ssh-add

paetling:
meant as a fix for issue #366 review @thieman

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -22,15 +22,11 @@ able to clone your private GitHub repos, you need to configure your standard use
 your `mac_username` is set to), to be able to clone repositories without any prompt for
 confirmation or password.
 
-See [this GitHub help article](https://help.github.com/articles/generating-ssh-keys/)
-for some info about setting up GitHub ssh keys.
-
-Since the daemon has no way to accept a password via user input, if your key requires a
-passphrase, you should run
 ```
 ssh-add -K <path-of-private-key>
 ```
-This will securely save the passphrase for that key in your Keychain.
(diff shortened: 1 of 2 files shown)


docs/faq.md
```diff
@@ -16,21 +16,19 @@ The Dusty daemon, which runs as root, manages your GitHub repos by default.
 To get permissions to clone from GitHub, the
 Daemon will make use of your unprivileged user's SSH_AUTH_SOCK. This means that the daemon's
 behavior using git should be the same as your user's, in terms of SSH authentication.
+(See [this GitHub help article](https://help.github.com/articles/generating-ssh-keys/)
+for some info about setting up GitHub ssh keys.)
```
thieman: I think we shouldn't actually send people here. Running an additional ssh-agent like it suggests fucks with things, plus it doesn't tell you to use `-K`.

docs/installation.md
```diff
@@ -17,14 +17,21 @@ programs:
...
 ```
 brew update
 which nginx || brew install nginx
 which boot2docker || brew install boot2docker
 which docker-compose || brew install docker-compose
 ```
+```
+ssh-add -K
+  -or-
+ssh-add -K <path-of-private-key>
```
thieman: Should say a little bit here about why we need this / which private key we need.

docs/faq.md
paetling: you are right.  This would cause confusion.  Should have looked. will remove.  talking like an idiot here. cannot stop. help. me. i. am. lost

paetling:
will squash my commits before I merge.

PR #410, opened 2015-07-14
cut version 0.2.2

jsingle:
@thieman cc @paetling @zacharym

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,8 @@
 # Changelog
 
-## 0.2.2 (In Progress)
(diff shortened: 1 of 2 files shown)


dusty/constants.py
```diff
@@ -1,6 +1,6 @@
 import os
 
-VERSION = '0.2.2'
+VERSION = '0.2.3'
 BINARY = False # overridden by PyInstaller when we build a binary
```
paetling: This should not happen till after the release

zacharym:
what alex said, otherwise looks good

PR #412, opened 2015-07-14
Fix test flap (hopefully)

jsingle:
@zacharym that error you saw happened again... it's transient but I hope this will decrease the frequency


zacharym:
:ghost: :cookie:

PR #420, opened 2015-07-24
Add a note about destructive rsync

thieman:
@jsingle


jsingle:
Worth adding a note for the cli usage too?

PR #423, opened 2015-07-24
Remove nginx dependency!

### LINKED ISSUE #393: Run nginx in a container instead of on the host
This would be a big gain for usability and simplicity on getting started.

thieman:
@jsingle https://www.youtube.com/watch?v=GGXzlRoNtHU

I've got it working, gonna stress test it a bit and make sure all the docs and such get updated properly.

Closes #393.


thieman:
image

dusty/constants.py
```diff
@@ -27,12 +25,17 @@
 DUSTY_GITHUB_PATH = 'gamechanger/dusty'
 DUSTY_BINARY_NAME = 'dusty'
 
-NGINX_MAX_FILE_SIZE = "500M"
+VM_IP_FROM_DOCKER = '172.17.42.1'
```
jsingle: This is a magic IP that docker containers always have?

thieman: This should always be the IP of the VM within the Docker bridge network. I think.

tests/unit/compiler/nginx/init_test.py
```diff
@@ -49,25 +46,22 @@ def test_get_nginx_configuration_spec_2(self):
                     proxy_set_header Connection "upgrade";
                     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
                     proxy_set_header Host $host;
-                    proxy_pass http://127.0.0.0:8000;
+                    proxy_pass http://172.17.42.1:8000;
```
jsingle: We could just use the constant for these yeah?

dusty/constants.py
jsingle: Is it something that boot2docker specifically sets up?

thieman: Nah it's how Docker works

tests/unit/compiler/nginx/init_test.py
thieman: Generally try to avoid using variables in the expected test output, this way you're actually writing down what you expect to see.

jsingle:
Do we need to expose any visibility as to the state of the nginx container?

thieman:
Probably a good idea to add it to status and warn if it dies. Will do those as separate PRs on top of the main functionality here.

jsingle:
This is pretty cool, go docker

thieman:
TESTS PASSIN YO

jsingle:
:boom: :trophy: :boom:

thieman:
Alright gonna merge this, get a RC cut, and try to get this out to the GC friends for testing

dusty/compiler/nginx/__init__.py
```diff
@@ -1,7 +1,7 @@
-from ...constants import NGINX_MAX_FILE_SIZE
+from ...constants import NGINX_MAX_FILE_SIZE, VM_IP_FROM_DOCKER
 
 def _nginx_proxy_string(port_spec):
-    return "proxy_pass http://{}:{};".format(port_spec['boot2docker_ip'], port_spec['proxied_port'])
+    return "proxy_pass http://{}:{};".format(VM_IP_FROM_DOCKER, port_spec['proxied_port'])
```
paetling: why is this better?

paetling: so nginx lives inside a container how is this getting out to the virtual machine?  Or is this, the ip of the outside world if I am in a container? if so that is is pretty nifty?

dusty/systems/nginx/__init__.py
```diff
@@ -37,5 +20,6 @@ def update_nginx_from_config(nginx_config):
     or tell it to reload its config to pick up what we've
     just written."""
     logging.info('Updating nginx with new Dusty config')
-    _write_nginx_config(nginx_config)
-    _ensure_nginx_running_with_latest_config()
+    temp_path = tempfile.mkstemp()[1]
+    _write_nginx_config(nginx_config, temp_path)
+    sync_local_path_to_vm(temp_path, os.path.join(constants.NGINX_CONFIG_DIR_IN_VM, 'dustyNginx.conf'))
```
paetling: why is this camel case?

thieman: I think I was intending to correlate it with the name of the service. I didn't want to call it `nginx.conf` because it isn't the main config file, it gets included from the main one (which is called `nginx.conf`).

paetling: This does not get included from the main one. This overwrites a location that nginx looks for its config files `/etc/nginx/conf.d`

paetling:
This is so cool!

dusty/systems/nginx/__init__.py
thieman: Fairly certain the main config is still at `/etc/nginx/nginx.conf` and it has an include to source `*.conf` from `/etc/nginx/conf.d`. Not in front of a Mac to confirm at the moment.

dusty/compiler/nginx/__init__.py
thieman: It's the IP of the VM (like, non-containerized VM TCL host) if you're a container. I should really draw up a diagram of how this works now and throw it in the docs.

The Nginx and target containers both forward ports to the VM, then we do this to make it all tie together:

Mac -> VM -> Nginx container -> VM again -> Target container

PR #426, opened 2015-07-24
Go back to sourcing rsync from TCL repo

### LINKED ISSUE #406: Potentially remove the 64-bit rsync binary download
It seems like the `rsync` package may have made it to the 64-bit repository. If so, we should stop using our bespoke binary. https://github.com/boot2docker/boot2docker/issues/936#issuecomment-120992953

thieman:
@jsingle rsync has landed in the 64-bit TCL repo, which means we can get rid of our bespoke binary. Hooray! Closes #406.

### DIFF
--- a/dusty/systems/virtualbox/__init__.py
+++ b/dusty/systems/virtualbox/__init__.py
@@ -10,11 +10,7 @@ from ...log import log_to_client
 
 def _ensure_rsync_is_installed():
     logging.info('Installing rsync inside the Docker VM')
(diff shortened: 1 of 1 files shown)


jsingle:
WOOOO!!!!!!!!

thieman:
Actually crap I probably need to do something to make sure things don't break if you already had the old thing installed. Maybe just keep the `which` check in place.

thieman:
Nevermind that last comment, it works fine if the old rsync is already there. Still failing tests though. :(

PR #428, opened 2015-07-24
Trigger disk space warnings when doing dusty up

### LINKED ISSUE #417: Rsync fails with exit status 12 when VM out of space
Report from @nrschultz. If the boot2docker VM is out of space and Dusty tries to sync, it'll fail with this really cryptic error message.

```
ERROR: Command '['rsync', '-e', 'ssh -p 2022 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i /Users/nick/.ssh/id_boot2docker', '-az', '--force', '--rsync-path', 'sudo rsync', '--exclude', '*/.git', '/gc/gcweb/', 'docker@localhost:/persist/repos/github.com/gamechanger/gcweb']' returned non-zero exit status 12
```

There's not much we can do about the error message specifically, but we should add some sort of user-facing warnings when the VM is running low on space so users can more quickly reason about what's going wrong.

thieman:
@jsingle Closes #417.

### DIFF
--- a/dusty/commands/run.py
+++ b/dusty/commands/run.py
@@ -11,6 +11,7 @@ from .. import constants
 from ..command_file import make_up_command_files
 from ..source import Repo
 from ..payload import daemon_command
+from ..warnings import daemon_warnings
 
 @daemon_command
 def start_local_env(recreate_containers=True, pull_repos=True):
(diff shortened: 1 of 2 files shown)


dusty/commands/run.py
```diff
@@ -36,6 +37,13 @@ def start_local_env(recreate_containers=True, pull_repos=True):
             log_to_client("WARNING: docker-compose stop failed")
             log_to_client(str(e))
 
+    daemon_warnings.clear_namespace('disk')
+    df_info = virtualbox.get_docker_vm_disk_info(as_dict=True)
+    if 'M' in df_info['free'] or 'K' in df_info['free']:
```
thieman: Such engineering

jsingle:
Works for me :+1:

PR #429, opened 2015-07-24
Kill version warnings

thieman:
@jsingle These were a noble idea, but in practice they alert all the time so people are going to just ignore them. We can easily bring them back if we find a good use for them, e.g. you must have at least X version or Dusty will for sure break.


jsingle:
I think it's ok to kill them.  Not convinced that removing this functionality completely is the best strategy but don't feel strongly

Issue #425, opened 2015-07-24
Add introspection around state of the nginx container


thieman:
Probably want this to show up in `dusty status` and warn if it dies, since all host forwarding will stop working.

thieman:
Going to punt on the warning aspect of this for now, we'll see if this comes up as a real problem.

PR #434, opened 2015-07-28
Add nginx container to dusty status

thieman:
@paetling Part of #425. Also makes this way faster by not instantiating the client N times.


paetling:
looks good. :ghost: :cow:

PR #438, opened 2015-07-28
Fix #437 - Add timestamps to dusty logs command

### LINKED ISSUE #437: dusty logs should have an option for showing timestamps
Request from @melkorm. The Docker logs API has an option for timestamps, we should expose that through `dusty logs`. We should consider making that the default behavior, as well.

melkorm:
Hey guys, this is my first pull request to python project, so help will be required :)

Few things:
 - the test have been fixed as in issue #437 and default set to True
 - new test is written for disabling timestamps
 - wasn't able to build the bins to actually test it, new guide / help for this would be welcomed

Enjoy :)

### DIFF
--- a/dusty/commands/logs.py
+++ b/dusty/commands/logs.py
@@ -3,7 +3,7 @@ import logging
 from ..systems.docker import get_dusty_containers
 from . import utils
 
-def tail_container_logs(app_or_service_name, follow=False, lines=None):
+def tail_container_logs(app_or_service_name, follow=False, lines=None, timestamps=True):
     containers = get_dusty_containers([app_or_service_name], include_exited=True)
     if len(containers) == 0:
         logging.info('No container exists which corresponds to {}'.format(app_or_service_name))
@@ -12,6 +12,8 @@ def tail_container_logs(app_or_service_name, follow=False, lines=None):
     args = ['logs']
     if follow:
         args.append('-f')
+    if timestamps:
+        args.append('-t')
     if lines:
         args.append('--tail={}'.format(lines))
     args.append(container['Id'])
--- a/dusty/commands/logs.py
+++ b/dusty/commands/logs.py
@@ -3,7 +3,7 @@ import logging
 from ..systems.docker import get_dusty_containers
 from . import utils
 
-def tail_container_logs(app_or_service_name, follow=False, lines=None, timestamps=True):
+def tail_container_logs(app_or_service_name, follow=False, lines=None, timestamps=False):
     containers = get_dusty_containers([app_or_service_name], include_exited=True)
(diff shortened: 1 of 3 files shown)


paetling:
@melkorm what do you mean you were not able to build bins to actually test it?   The test you added can be run using nosetests from the dusty repo

thieman:
+testonce

dusty/commands/logs.py
```diff
@@ -3,7 +3,7 @@
 from ..systems.docker import get_dusty_containers
 from . import utils
 
-def tail_container_logs(app_or_service_name, follow=False, lines=None):
+def tail_container_logs(app_or_service_name, follow=False, lines=None, timestamps=True):
```
thieman: There's a disconnect here between the CLI and this function (which is only called by the CLI). In the CLI itself, timestamps default to off (which, now that I think about it, should be how it works to mimic Docker itself), but in this function they're defaulted to True.

Think we just need to default this to False here then adjust the tests back.

thieman:
@melkorm Looks great! One thing about changing the defaults, then we can test it again and get it merged. I'll probably write an integration test for it after that and CC you on it.

melkorm:
@paetling I did run the tests with nosetests - wouldn't do the PR other way. Was not really sure how to run integration ones / build bins to run it, and run logs command directly on this build.
This line is the most confusing for me:
`$ sudo dusty -d & # launch a daemon based on your checked out code` - how can I build the bin ?

@thieman Yes, I did it with False at first, but then I saw the issue report hich said to set it to True, so I did so - I will revert it back to False.

thieman:
+testonce

You're right, this can all be a hell of a lot clearer, I'll try to write some better documentation and throw it in the Contributing section of the docs.

For now, I think you should be able to massage these commands to run your integration tests:

```bash
# stop your installed Dusty daemon
sudo launchctl stop com.gamechanger.dusty 
# make a virtualenv for Dusty
virtualenv venv 
# install Dusty and its requirements into the virtualenv
python setup.py develop 
# launch a Dusty daemon, which'll work in your virtualenv
sudo dusty -d & 
# then actually run the integration tests, replacing yourusername with your
# actual username on your mac
sudo DUSTY_ALLOW_INTEGRATION_TESTS=yes DUSTY_INTEGRATION_TESTS_USER=yourusername nosetests -v tests/integration
```

It's probably worth refining this a bit and then adding a script to the repo itself to make this easier for people.

thieman:
+testonce

thieman:
@melkorm I'm going to go ahead and merge this so we can get it into the next release, will follow it up shortly with another PR. We're making progress on making integration tests easier to run, as well.

PR #441, opened 2015-07-29
Explicitly state that mount is the location once and always scripts start in 

paetling:
review @thieman  
image


thieman:
:bear:

PR #443, opened 2015-07-30
Customize socket location through DUSTY_SOCKET_PATH

thieman:
@jsingle Don't think this is particularly worth integration testing, thoughts?


jsingle:
I agree not worth an integration test :+1:

PR #444, opened 2015-07-30
Js http cloning

jsingle:
@thieman cloning w https is easy; might make it easier for some people to get set up...


thieman:
:bear:

PR #445, opened 2015-07-30
Integration test and changelog for log timestamps

thieman:
Review @jsingle cc @melkorm


jsingle:
Is this fixed in the "usage" documentation?

thieman:
Good call.

PR #448, opened 2015-07-30
Add script for easier integration testing

### LINKED ISSUE #440: Write a script for easier integration testing locally
Blocked on #439. Currently there are a few steps and environment variables you need to set for the integration tests to work nicely, we should be able to automate most of that through a simple script.

thieman:
@jsingle Fixes #440. Need to write some docs to go with it still.

### DIFF
--- a/tests/run_integration_tests.sh
+++ b/tests/run_integration_tests.sh
@@ -25,6 +25,11 @@ ROOT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && cd .. && pwd )
 TEMP_REQUIREMENTS_PATH=$TMPDIR"dusty-test-requirements.txt"
 TEST_USER=`whoami`
 
+TEST_PATH=$ROOT_DIR/tests/integration
+if [ ! -z "$1" ]; then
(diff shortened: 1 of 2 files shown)


tests/run_integration_tests.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+
+# Check that the script is being run correctly and can find the Dusty executable
+
+if [ -z ${VIRTUAL_ENV+x} ]; then
```
jsingle: crazy

jsingle:
Pretty cool if this works.  You gonna change the jenkins job to use this?

thieman:
Eh if it ain't broke

jsingle:
I feel like we should have one way to run this stuff.  That way we'll be forced to keep this script up to date

jsingle:
Yeah after looking more at the jenkins script and this one I agree it's not worth consolidating them

jsingle:
Looks good to me.  I think it's OK that we're leaving upgrade integration tests out, although it shouldn't be hard to add support

Issue #449, opened 2015-07-30
Dusty is reliant on `mount` in the app spec


paetling:
Right now, on line #67 of dusty/command_file.py we have a `cd {}` command we are adding.   If you do not specify a mount or a repo, this command is still executed as just `cd`.   This can lead to errors on startup for apps that do not have a `mount` key.

PR #450, opened 2015-07-30
Remove mount dependency

paetling:
review @jsingle  do not cd or export path with that directory if mount is not specified.  closes issue #449


dusty/command_file.py
```diff
@@ -13,6 +13,8 @@ def _write_commands_to_file(list_of_commands, file_location):
         os.makedirs(file_location_parent)
     with open(file_location, 'w+') as f:
         for command in list_of_commands:
+            import logging
+            logging.error(command)
```
paetling: removing

jsingle:
:tada:

Issue #436, opened 2015-07-28
App schema: `commands` is not required


thieman:
Report from investigation on #435. An app spec will pass validation if it has no `commands` key at all. If it does have one, however, the `always` key must be defined. I think the intention is that `commands: always` must always be defined.

thieman:
Fixed by #436.

PR #451, opened 2015-07-30
Commands key should be required

jsingle:
@paetling see https://github.com/gamechanger/dusty/issues/436


paetling:
your tests failed but this seems like a good idea.  My guess is that are tests are not specifying commands in places which is now not legal

jsingle:
Yeah I have to manually change 10000 different dicts to add these keys :rage:

PR #457, opened 2015-07-30
Force Docker to use our commands, not any defined ENTRYPOINT

### LINKED ISSUE #435: Images with ENTRYPOINT defined don't work as apps
Report originally from Slack by @melkorm. Images with an `ENTRYPOINT` defined don't play nicely with Dusty's attempts to register the command file as a `COMMAND`, since this'll end up passing the command file to an entrypoint binary that has no idea what to do with it.

This app spec is a workaround to the problem with an image (in this case, `berngp/docker-zabbix`) which has an ENTRYPOINT defined. If you remove the `compose: entrypoint` key, you will see the error. This key is a workaround for now.

```yaml
image: berngp/docker-zabbix

host_forwarding:
  - host_name: local.zabbix.com
    host_port: 80
    container_port: 80

commands:
  always:
    - /bin/docker-zabbix run

compose:
  entrypoint: []
```

I think the solution here is to enforce the `commands` pattern for apps and always blow away the entrypoint if it is defined.

thieman:
@paetling Fixes #435. I confirmed that @melkorm's Zabbix example works fine with this fix without having the workaround explicitly defined.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## 0.4.0 (In Progress)
- * BREAKING CHANGE: Dusty now cleans up volumes when removing containers
+ * BREAKING CHANGE: Dusty now removes data volumes when removing containers. Containers needing to persist data should use volume mounts inside of `/persist` in the boot2docker VM.
 
  * NEW: Dusty no longer requires nginx to be installed on your Mac! Dusty now runs a containerized nginx inside Docker instead. All other functionality around host forwarding is unchanged.
  * NEW: Dusty's socket location can now be customized via the `DUSTY_SOCKET_PATH` environment variable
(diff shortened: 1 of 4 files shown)


docs/changelog.md
```diff
@@ -1,7 +1,7 @@
 # Changelog
 
 ## 0.4.0 (In Progress)
- * BREAKING CHANGE: Dusty now cleans up volumes when removing containers
+ * BREAKING CHANGE: Dusty now removes data volumes when removing containers. Containers needing to persist data should use volume mounts inside of `/persist` in the boot2docker VM.
```
paetling: where is this in your code?

thieman: I'm touching up the changelog message that was already there. It's a drive-by.

docs/changelog.md
```diff
@@ -9,6 +9,7 @@
  * NEW: `dusty logs` now supports a `-t` option to show timestamps on the logging output
 
  * FIXED: `dusty upgrade` will now successfully upgrade Dusty from a RC version in all cases
+ * FIXED: Images with an `ENTRYPOINT` defined now work correctly when used as the basis for a Dusty app
```
paetling: We should add to the docs that ENTRYPOINTS in Dockerfiles are ignored

thieman: Smart

PR #460, opened 2015-07-30
Js colons fix for volumes

jsingle:
@thieman Docker Compose parses `volumes` by splitting on colons.  Having colons in a local path will mess that up

Fixes https://github.com/gamechanger/dusty/issues/459

### DIFF
--- a/dusty/config.py
+++ b/dusty/config.py
@@ -47,8 +47,16 @@ def save_config_value(key, value):
         verify_mac_username(value)
         save_config(current_config)
         check_and_load_ssh_auth()
+    elif key == constants.CONFIG_REPO_OVERRIDES_KEY:
+        verify_overrides_without_colons(current_config)
+        save_config(current_config)
     else:
         save_config(current_config)
+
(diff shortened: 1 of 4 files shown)


thieman:
There's no way around this? Is there an issue in Compose?

jsingle:
I haven't found anything.  If you think this is really messy I can look a bit more, but I don't think this is that bad...

thieman:
Uuuuuuugh ok. It's just a lot of janky code to deal with a corner case. How did we even find this bug? Are people actually using colons in repo names / directories? Those people are dumb

jsingle:
I found it because when you specify https://github/blah, the local path contains a folder called https://github/blah, and docker-compose breaks.  So at the very least, we need to handle the case where people enter ssh: or http: as part of the path.  There might be a slightly cleaner way to do that particular thing, but the colons are allowed in filepaths anyway, so I thought I might as well make a more general check

thieman:
:bear:

PR #476, opened 2015-08-07
split out config file module

jsingle:
@thieman we can use this new module for `/etc/exports` as well as hosts, yeah?

### DIFF
--- a/dusty/constants.py
+++ b/dusty/constants.py
@@ -1,10 +1,15 @@
 import os
+import re
 
 VERSION = '0.4.0'
 BINARY = False # overridden by PyInstaller when we build a binary
 PRERELEASE = False # overridden by PyInstaller when we build a prerelease binary
 
 EXAMPLE_SPECS_REPO = 'github.com/gamechanger/dusty-example-specs'
+
+DUSTY_CONFIG_BEGIN = '# BEGIN section for Dusty\n'
+DUSTY_CONFIG_END = '# END section for Dusty\n'
+DUSTY_CONFIG_REGEX = re.compile('\\{}.*\\{}'.format(DUSTY_CONFIG_BEGIN, DUSTY_CONFIG_END), flags=re.DOTALL | re.MULTILINE)
 
 SOCKET_TERMINATOR = '\0\0'
 SOCKET_ERROR_TERMINATOR = '\0\1'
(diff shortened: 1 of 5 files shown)


dusty/constants.py
```diff
@@ -6,6 +7,10 @@
 
 EXAMPLE_SPECS_REPO = 'github.com/gamechanger/dusty-example-specs'
 
+DUSTY_CONFIG_BEGIN = '# BEGIN section for Dusty\n'
+DUSTY_CONFIG_END = '# END section for Dusty\n'
+DUSTY_CONFIG_REGEX = re.compile('\\{}.*\\{}'.format(DUSTY_CONFIG_BEGIN, DUSTY_CONFIG_END), flags=re.DOTALL | re.MULTILINE)
```
thieman: The regex is a touch different than before since begin has a `\n` now, maybe make sure it's all still kosher?

dusty/systems/hosts/__init__.py
```diff
@@ -1,34 +1,17 @@
...
-    return DUSTY_CONFIG_REGEX.sub("", hosts)
 
 def _dusty_hosts_config(hosts_specs):
     """Return a string of all host rules required to match
     the given spec. This string is wrapped in the Dusty hosts
     header and footer so it can be easily removed later."""
-    rules = '# BEGIN section for Dusty\n'
+    rules = ''
     for spec in hosts_specs:
         rules += '{} {}\n'.format(spec['forwarded_ip'], spec['host_address'])
```
thieman: This can just be a list comprehension now I think

jsingle: yep

dusty/constants.py
jsingle: Unittests indicate that things are OK...

tests/unit/systems/config_file/init_test.py
```diff
@@ -0,0 +1,32 @@
...
+        """)
+        self.non_spec_starter = "127.0.0.1 some-host.local.com\n"
+
+    def test_remove_current_dusty_config_from_blank(self):
+        result = config_file.remove_current_dusty_config(self.spec_output)
+        self.assertEqual(result, "")
+
+    def test_remove_current_dusty_config_from_starting(self):
+        result = config_file.remove_current_dusty_config(self.non_spec_starter + self.spec_output)
+        self.assertEqual(result, self.non_spec_starter)
```
thieman: Probably worth making sure that stuff after the Dusty section also doesn't get blown away

PR #480, opened 2015-08-11
Automatically apply fix for virtio networking

### LINKED ISSUE #479: Implement fix for VBox virtio networking issue
See discussion here https://github.com/boot2docker/boot2docker/issues/1022

We should build the fix for this into Dusty itself so we can get this out as quickly as possible. It should be sufficient to change the network interfaces to using PCnet-FAST III instead of virtio.

This is blocking #442.

thieman:
@paetling cc @jsingle Closes #479.

### DIFF
--- a/dusty/systems/virtualbox/__init__.py
+++ b/dusty/systems/virtualbox/__init__.py
@@ -1,7 +1,11 @@
+from __future__ import absolute_import
+
 import os
 import re
 import logging
 import textwrap
+import time
+from subprocess import CalledProcessError
 
 from ... import constants
 from ...config import get_config_value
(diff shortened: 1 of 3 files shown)


thieman:
Will add a note to changelog on this as well.

dusty/systems/virtualbox/__init__.py
```diff
@@ -40,8 +40,40 @@ def _start_docker_vm():
...
+    for line in _get_vm_config():
+        if 'nictype' in line and 'virtio' in line:
+            return True
+    return False
+
+def _apply_virtio_networking_fix():
+    """Modify all boot2docker VM network interfaces to use the preferred VM NIC type (generally
+    in place of virtio)."""
+    for idx, _ in enumerate(filter(lambda line: 'nictype' in line, _get_vm_config())):
+        logging.info('Setting VM NIC #{} to {}'.format(idx + 1, constants.VM_NIC_TYPE))
```
paetling: these are guaranteed ordered? or it does not really matter since we will be hitting 1, 2, 3... all respective numbers anyway?

paetling:
why are we using this network Am79C973?

dusty/systems/virtualbox/__init__.py
thieman: I think you could theoretically set them up not starting at number 1, but that's not the default and I have no idea why anyone would do that.

thieman:
Ugh failing because `boot2docker stop` has a hard-coded limit of 10 seconds after which it just gives up UGH

thieman:
Also that's the code for PCnet-FAST III, which was the option that seemed to work the best for us in Jordan's testing

paetling:
we could do the same thing as memory changes.  Where we change it right before we start boot2docker

thieman:
You can't change the NIC type on a running VM, but you can with memory.

paetling:
Can you change the memory while it is running?  I did not think you could.  Either way you could attempt to change it before the vm starts.

thieman:
Actually yeah you can't change it while it's running, so now I'm wondering why the code we have doesn't break constantly.

thieman:
Ok it does fail, we're just ignoring the failure. I'll open a separate issue to shore up the memory handling and put it into a more rigorous check, then it can also take advantage of the stopped VM to set itself correctly.

PR #482, opened 2015-08-12
Js config file

jsingle:
@thieman this is a regex that returns the contents of the dusty config section

### DIFF
--- a/dusty/systems/config_file/__init__.py
+++ b/dusty/systems/config_file/__init__.py
@@ -1,9 +1,16 @@
+import logging
+import re
+
 from ... import constants
 
 
 def read(filepath):
-    with open(filepath, 'r') as f:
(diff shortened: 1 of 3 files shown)


dusty/systems/config_file/__init__.py
```diff
@@ -1,9 +1,16 @@
...
 
 def read(filepath):
-    with open(filepath, 'r') as f:
-        return f.read()
+    try:
+        with open(filepath, 'r') as f:
+            return f.read()
+    except IOError:
+        logging.info('Unable to read file {}'.format(filepath))
+        return ''
```
thieman: Exception handling like this belongs in the caller.

jsingle: OK that's reasonable

PR #484, opened 2015-08-12
nfs for repos

jsingle:
@thieman still needs tests, and some old rsync stuff needs to be ripped out.  This is ready for your input tho


dusty/systems/nfs/client.py
```diff
@@ -0,0 +1,58 @@
...
+    mount_string = '-t nfs ' + _nfs_options_string()
+    mount_string += '{}:{} '.format(_host_ip(),repo.local_path)
+    mount_string += repo.vm_path
+    return mount_string
+
+def _nfs_options_string():
+    return '-o async,udp,noatime '
+
+def _host_ip():
+    return '192.168.59.3'
```
jsingle: On my radar to try making this better.  Maybe a separate PR though, don't think it's simple

thieman: Is this guaranteed?

dusty/systems/nfs/client.py
```diff
@@ -0,0 +1,58 @@
...
+    return check_output_demoted(['boot2docker', 'ssh', 'sudo mount {}'.format(_nfs_mount_args_string(repo))])
+
+def _nfs_mount_args_string(repo):
+    mount_string = '-t nfs ' + _nfs_options_string()
+    mount_string += '{}:{} '.format(_host_ip(),repo.local_path)
+    mount_string += repo.vm_path
+    return mount_string
+
+def _nfs_options_string():
+    return '-o async,udp,noatime '
```
thieman: Is the space at the end of this necessary? If so, it belongs in a format string in the caller, not in here. Confusing.

dusty/systems/nfs/client.py
```diff
@@ -0,0 +1,58 @@
...
+        _mount_repo(repo, wait_for_server=wait_for_server)
+
+def _start_nfs_client():
+    check_call_demoted(['boot2docker', 'ssh', 'sudo /usr/local/etc/init.d/nfs-client start'])
+
+def _unmount_repo(repo):
+    call_demoted(['boot2docker', 'ssh', 'sudo umount -l {}'.format(repo.vm_path)])
+
+def _mount_repo(repo, wait_for_server=False):
+    check_call_demoted(['boot2docker', 'ssh', 'sudo mkdir -p {}'.format(repo.vm_path)])
```
thieman: Docstring on this hairy-lookin function please

dusty/systems/nfs/__init__.py
```diff
@@ -0,0 +1,10 @@
+from . import client
+from . import server
+
+def configure_nfs():
+    server.configure_nfs_server()
+    client.mount_active_repos()
+
+def update_nfs_with_repos(repos):
+    server.add_exports_for_repos(repos)
+    client.remount_repos(repos)
```
thieman: Is this assuming the server is up and running? We may need to add explicit checks for that and set it up if not. Otherwise the implicit dependency will cause bugs or complication in other parts of the code.

thieman: For an example scenario, `dusty test` has no dependency on `dusty up` having been run.

dusty/commands/run.py
```diff
@@ -93,10 +93,10 @@ def restart_apps_or_services(app_or_service_names=None, sync=True):
         if app_or_service_names:
             specs = spec_assembler.get_assembled_specs()
             specs_list = [specs['apps'][app_name] for app_name in app_or_service_names if app_name in specs['apps']]
-            rsync.sync_repos_by_specs(specs_list)
+            repos_list = [spec['repo'] for spec in specs_list if spec['repo']]
+            nfs.update_nfs_with_repos(repos_list)
         else:
-            rsync.sync_repos(spec_assembler.get_all_repos(active_only=True, include_specs_repo=False))
-
+            nfs.update_nfs_with_repos(spec_assembler.get_all_repos(active_only=True, include_specs_repo=False))
```
thieman: I think it might make sense to move all of the NFS handling shit to `up` and `test`. We can talk about it. Basically, if you do anything during restart, you're going to yank NFS mounts away from running processes, which seems like a recipe for weirdness.

dusty/systems/nfs/client.py
```diff
@@ -0,0 +1,58 @@
...
+from ...subprocess import call_demoted, check_call_demoted, check_output_demoted
+from ...compiler.spec_assembler import get_all_repos
+
+def mount_active_repos():
+    remount_repos(get_all_repos(active_only=True, include_specs_repo=False))
+
+def remount_repos(repos):
+    _start_nfs_client()
+    for i, repo in enumerate(repos):
+        wait_for_server = not i
```
thieman: lol wat

dusty/systems/nfs/__init__.py
jsingle: You're right we need extra checks for a running server

dusty/commands/run.py
jsingle: Yeah we can talk.  Anecdotally the containers recover fine in this scenario

dusty/systems/nfs/client.py
jsingle: Only the first mount command should be waiting for the server to restart

dusty/systems/nfs/client.py
thieman: Praise science `vboxmanage list hostonlyifs`. But yeah not a blocker. Let's make sure we do fix it at some point.

dusty/systems/nfs/client.py
```diff
@@ -0,0 +1,58 @@
...
+                    time.sleep(1)
+                else:
+                    raise e
+        log_to_client('Failed to mount repo {}'.format(repo.short_name))
+        raise RuntimeError('Unable to mount repo with NFS')
+    else:
+        _run_mount_command(repo)
+
+def _run_mount_command(repo):
+    return check_output_demoted(['boot2docker', 'ssh', 'sudo mount {}'.format(_nfs_mount_args_string(repo))])
```
thieman: Should this be `check_call`?

dusty/systems/nfs/client.py
thieman: Could we do something like this instead?

`_mount_repo(repo, wait_for_server=True if i==0 else False)`

dusty/systems/nfs/client.py
jsingle: Hmm that looks a little better than the python fuckery i was considering

dusty/systems/nfs/client.py
jsingle: haha sure

dusty/systems/nfs/client.py
jsingle: I think i might need `check_output`, so that when an error is raised, I can check `e.output`

dusty/systems/nfs/server.py
```diff
@@ -0,0 +1,91 @@
...
+def _export_for_repo(repo, vm_ip):
+    return '{} {} -alldirs -maproot={}\n'.format(os.path.realpath(repo.local_path), vm_ip, _maproot_for_repo(repo))
+
+def _maproot_for_repo(repo):
+    stat = os.stat(repo.local_path)
+    return '{}:{}'.format(stat.st_uid, stat.st_gid)
+
+def _restart_server():
+    try:
+        check_output(['nfsd', 'update'], demote=False)
```
thieman: Let's make a note to try to get Mavericks and The Capitan testers for this RC

dusty/systems/nfs/server.py
```diff
@@ -0,0 +1,91 @@
...
+    except CalledProcessError as e:
+        if 'nfsd not running' in e.output:
```
thieman: Do we have any better ways of testing this? Can we use `psutil` or something to directly check whether nfsd is running?

dusty/systems/nfs/server.py
jsingle: done

dusty/systems/nfs/server.py
```diff
@@ -0,0 +1,91 @@
...
+            log_to_client('Restarting NFS Server')
+            check_and_log_output_and_error(['nfsd', 'restart'], demote=False)
+        else:
+            raise e
+
+def _exports_set():
```
thieman: Naming on this function is kinda weird

dusty/systems/nfs/server.py
```diff
@@ -0,0 +1,91 @@
...
+
+def _export_for_dusty_managed(vm_ip):
+    return '{} {} -alldirs -maproot=0:0\n'.format(os.path.realpath(constants.REPOS_DIR), vm_ip)
+
+def _export_for_repo(repo, vm_ip):
+    return '{} {} -alldirs -maproot={}\n'.format(os.path.realpath(repo.local_path), vm_ip, _maproot_for_repo(repo))
+
+def _maproot_for_repo(repo):
+    stat = os.stat(repo.local_path)
+    return '{}:{}'.format(stat.st_uid, stat.st_gid)
```
thieman: If I change permissions on a folder, it'll get picked up for a reconfig the next time I do `dusty up` or whatever, yeah?

thieman:
Writing integration tests for this one is gonna be fun

thieman:
Also Jordan I am SO STOKED FOR THIS

dusty/systems/nfs/server.py
jsingle: Yeah that should happen

dusty/systems/nfs/client.py
thieman: Ah, ok. This is brittle as-is, the caller's dependency on this is implicit. At a minimum, you should add a comment explaining why it needs to stay `output`, and if possible you should refactor it to make the dependency explicit.

dusty/systems/nfs/server.py
jsingle: Is that more direct than checking `nfsd status` output?

thieman: It's more direct and less brittle than checking the output for a particular string.

dusty/commands/run.py
jsingle: Also being able to restart to swap out repos is pretty slick

jsingle:
praise NFS!

dusty/systems/nfs/server.py
```diff
@@ -0,0 +1,100 @@
...
+
+def _check_exports():
+    check_and_log_output_and_error(['nfsd', 'checkexports'])
```
thieman: I think it's likely that some of the more advanced users who will try us out may run into issues with this, particularly if they are already using an NFS setup with b2d/Machine for their local dev. I think it's worth trying to make the error message out of this as friendly as possible. Explain the problem, tell them how to fix it, etc etc

jsingle: Yeah I've added more logging locally.  I'll make it better tho

PR #485, opened 2015-08-13
Be more persistent about getting VM memory to configured value

### LINKED ISSUE #481: Try harder to set memory size for VM correctly
Currently, Dusty will only set the VM's memory size to the managed value if you happen to run a `dusty up` while the VM is powered off. We should be able to piggyback on some of the code introduced in #480   to check the memory value, take advantage of the code we already have to power down the VM, then set the memory value appropriately when it diverges.

thieman:
@jsingle Fixes #481.

### DIFF
--- a/dusty/commands/disk.py
+++ b/dusty/commands/disk.py
@@ -13,7 +13,6 @@ from ..payload import daemon_command
 
 @daemon_command
 def cleanup_inactive_containers():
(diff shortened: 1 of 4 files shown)


jsingle:
Nice

jsingle:
Integration tests might be broken because of nfs stuff?

PR #486, opened 2015-08-13
Fix command file issues on case-insensitive filesystems

### LINKED ISSUE #475: Commands fail after changing capitalization of spec on HFS
From @nrschultz. Due to the case-insensitive nature of the HFS filesystem used on OS X, changing the capitalization of the spec file of an app or lib can cause the container to crash following subsequent `dusty up` commands with an error message like the following:

```
sh: 0: Can't open /command_files/dusty_command_file_myApp.sh
```

In this example, if you changed the app spec at `myapp.yml` to instead be `myApp.yml`, you could see this error. This is because the command files written to `/etc/dusty/commands` on the user's local system (Mac) do not pick up the change in capitalization on case-insensitive filesystems.

"Quite frankly, HFS+ is probably the worst filesystem ever." - Linus Torvalds

thieman:
@jsingle cc @nrschultz Fixes #475.

How did anyone ever think case-insensitive filesystems were a good idea? :anguished:

### DIFF
--- a/tests/integration/cli/up_test.py
+++ b/tests/integration/cli/up_test.py
@@ -1,3 +1,4 @@
+import time
 import subprocess
 
 from ...testcases import DustyIntegrationTestCase
@@ -8,6 +9,13 @@ class TestUpCLI(DustyIntegrationTestCase):
         super(TestUpCLI, self).setUp()
         busybox_single_app_bundle_fixture(num_bundles=2)
         self.run_command('bundles activate busyboxa busyboxb')
+
+    def tearDown(self):
+        try:
+            self.run_command('stop')
+        except Exception:
+            pass
+        super(TestUpCLI, self).tearDown()
 
     def test_basic_up_command(self):
         run_output = self.run_command('up')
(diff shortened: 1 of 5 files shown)


tests/integration/cli/up_test.py
```diff
@@ -9,6 +10,13 @@ def setUp(self):
         busybox_single_app_bundle_fixture(num_bundles=2)
         self.run_command('bundles activate busyboxa busyboxb')
 
+    def tearDown(self):
+        try:
+            self.run_command('stop')
+        except Exception:
+            pass
+        super(TestUpCLI, self).tearDown()
```
thieman: Explicitly doing the stop makes the `up` integration tests much faster. Without this, we end up doing slow teardowns at the beginning of each subsequent test, which takes 10 seconds per container. With this, it's about 1 second per container.

jsingle: That's interesting  - why would `stop` be much faster?

tests/integration/cli/up_test.py
```diff
@@ -9,6 +10,13 @@ def setUp(self):
         busybox_single_app_bundle_fixture(num_bundles=2)
         self.run_command('bundles activate busyboxa busyboxb')
 
+    def tearDown(self):
+        try:
+            self.run_command('stop')
+        except Exception:
```
jsingle: except Exception :fuckyeah:

jsingle:
:pig:

tests/integration/cli/up_test.py
thieman: When we explicitly stop we're able to provide a timeout value on the SIGTERM. We pass this as 1 second, down from the default of 10 seconds.

If you do a `docker-compose up` and there are old versions of the containers running, it'll stop them for you with the 10 second default.

jsingle: oh gotcha

thieman:
I'm merging this. All integration tests are passing on my local, I think the NFS stuff is seriously fucking with the b2d VM in use on the Jenkins box.

thieman:
On second thought if this last run doesn't pass I'll leave this here and revisit it once NFS is stable. This isn't urgent.

PR #489, opened 2015-08-14
intelligently retrieve the host ip on the VM network

jsingle:
@thieman


jsingle:
Integration tests :imp:

PR #491, opened 2015-08-14
Js memoize

jsingle:
@thieman MEMOIZE

### DIFF
--- a/dusty/memoize.py
+++ b/dusty/memoize.py
@@ -1,0 +1,24 @@
+import functools
+import pickle
+
+from .payload import function_key
+
+cache = {}
+
+def memoized(fn):
+    """
+    Decorator. Caches a function's return value each time it is called.
+    If called later with the same arguments, the cached value is returned
+    (not reevaluated).  The cache lasts for the duration of each request.
+    """
+    @functools.wraps(fn)
+    def memoizer(*args, **kwargs):
+        key = function_key(fn) + pickle.dumps(args) + pickle.dumps(kwargs)
+        if key not in cache:
+            cache[key] = fn(*args, **kwargs)
+        return cache[key]
+    return memoizer
+
+def reset_memoize_cache():
+    global cache
+    cache = {}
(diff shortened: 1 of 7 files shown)


dusty/memoize.py
```diff
@@ -0,0 +1,24 @@
+import functools
+import pickle
+
+from .payload import function_key
+
+cache = {}
+
+def memoized(fn):
+    '''
```
thieman: """

tests/testcases.py
```diff
@@ -49,12 +50,14 @@ def setUp(self):
...
         self.capture_handler = TestCaptureHandler(self.client_output)
         logging.getLogger(constants.SOCKET_LOGGER_NAME).addHandler(self.capture_handler)
+        reset_memoize_cache()
 
     def tearDown(self):
         os.remove(self.temp_config_path)
         shutil.rmtree(self.temp_specs_path)
         shutil.rmtree(self.temp_repos_path)
         logging.getLogger(constants.SOCKET_LOGGER_NAME).removeHandler(self.capture_handler)
+        reset_memoize_cache()
```
thieman: Can't we get away with just doing this at the end of each tearDown, not also during setUp?

tests/unit/memoize_test.py
```diff
@@ -0,0 +1,24 @@
...
+    def setUp(self):
+        super(TestMemoize, self).setUp()
+        self.counter = 0
+        @memoized
+        def memoized_fn():
+            self.counter += 1
+        self.memoized_fn = memoized_fn
+
+    def test_memoize(self):
+        self.memoized_fn()
```
thieman: Probably worth adding a test to make sure the function is actually returning what you think it should be

tests/testcases.py
jsingle: nope :)

tests/unit/memoize_test.py
jsingle: +1

dusty/memoize.py
jsingle: Haha I WAS COPYING GCLIB

tests/testcases.py
thieman: :anguished:

jsingle: Yeah not sure why but that's what i observed. Didn't feel like digging into why

PR #495, opened 2015-08-14
hash kwargs

jsingle:
@thieman


jsingle:
If you think this is good, should we do this in gclib too?

thieman:
Arguably we should just grab a library for this, but yeah seems fine

jsingle:
Like what kind of library?

PR #505, opened 2015-08-17
add reset --hard for dusty up

jsingle:
@thieman this adds `reset --hard origin/master` on dusty up only. In other cases when a repo gets pulled, this would log a warning message to the client

### DIFF
--- a/dusty/source.py
+++ b/dusty/source.py
@@ -130,7 +130,21 @@ class Repo(object):
         with git_error_handling():
             git.Repo.clone_from(self.assemble_remote_path(), self.managed_path)
 
-    def update_local_repo(self):
+    def get_latest_commit(self):
+        repo = git.Repo(self.managed_path)
+        for ref in repo.refs:
+            if ref.path == 'refs/remotes/origin/master':
+                return ref.commit
+        return repo.remote().fetch()[0].commit
+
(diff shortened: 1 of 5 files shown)


dusty/source.py
```diff
@@ -140,3 +154,11 @@ def update_local_repo(self):
...
         with git_error_handling():
             managed_repo.remote().pull('master')
+        if not self.local_is_up_to_date():
+            if force:
+                with git_error_handling():
+                    managed_repo.git.reset('--hard', 'origin/master')
+            else:
+                log_to_client('WARNING: couldn\'t update {} because of local conflicts.'.format(repo.local_path)
+                              'A container may have modified files in the repos\'s directory'
+                              'Running `dusty up` should resolve this')
```
thieman: This is a symptom of the chronic problem that the user needs to fix. We should phrase the error message to reflect that, not give them a workaround that'll lead to them hitting the problem again and again and again.

thieman:
LGTM on design

dusty/source.py
jsingle: Sounds good

jsingle:
gitignores are for sure honored

jsingle:
oops, broke some other stuff

PR #507, opened 2015-08-17
fix repos upgrade integration

jsingle:
@zacharym these tests needed some fixin

### DIFF
--- a/tests/integration/cli/repos_test.py
+++ b/tests/integration/cli/repos_test.py
@@ -24,19 +24,19 @@ class TestReposCLI(DustyIntegrationTestCase):
         os.chmod(self.fake_override_dir, 0777)
         os.chmod(self.fake_from_dir, 0777)
 
-        self.old_repos_dir = constants.REPOS_DIR
-        constants.REPOS_DIR = self.temp_repos_dir
         self.fake_override_repo_location = os.path.join(self.fake_override_dir, 'fake-repo')
         self._set_up_fake_local_repo(path=self.fake_override_repo_location)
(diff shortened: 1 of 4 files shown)


jsingle:
OOps broke other stuff

zacharym:
going to wait for this to pass, then i'll take a look

tests/integration/cli/repos_test.py
```diff
@@ -24,19 +24,19 @@ def setUp(self):
         os.chmod(self.fake_override_dir, 0777)
         os.chmod(self.fake_from_dir, 0777)
 
-        self.old_repos_dir = constants.REPOS_DIR
-        constants.REPOS_DIR = self.temp_repos_dir
         self.fake_override_repo_location = os.path.join(self.fake_override_dir, 'fake-repo')
         self._set_up_fake_local_repo(path=self.fake_override_repo_location)
         self.fake_from_repo_location = os.path.join(self.fake_from_dir, 'fake-repo')
         self._set_up_fake_local_repo(path=self.fake_from_repo_location)
+        self.fake_source_repo_location = '/tmp/fake-repo'
```
zacharym: is there a reason why we are stuck using /tmp instead of mkdtemp ? I seem to recall there being something why

zacharym:
seems legit

tests/integration/cli/repos_test.py
jsingle: The code could in theory be better.  But the specs fixture needs to point to a defined place, not something that we can dynamically create in the testcase.  at least with the current pattern

PR #510, opened 2015-08-17
Manually sync before checking file existence

thieman:
@jsingle Let's see if this stops these file checks from failing


jsingle:
OK... you'll have to run this several times to be sure that it's fixed, since this is an intermittent failure

PR #511, opened 2015-08-17
More rigorous check for image-tag combo

thieman:
@jsingle Fixes a bug where you had one tag of the image locally, but not necessarily the tag you needed.

### DIFF
--- a/dusty/systems/docker/testing_image.py
+++ b/dusty/systems/docker/testing_image.py
@@ -41,8 +41,11 @@ def _get_create_container_binds(split_volumes):
     return binds_dict
 
 def _ensure_image_exists(docker_client, image_name):
+    full_image_name = image_name
+    if ':' not in image_name:
+        full_image_name = '{}:latest'.format(image_name)
     for image in docker_client.images():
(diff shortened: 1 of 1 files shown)


dusty/systems/docker/testing_image.py
```diff
@@ -41,8 +41,11 @@ def _get_create_container_binds(split_volumes):
...
 def _ensure_image_exists(docker_client, image_name):
+    full_image_name = image_name
+    if ':' not in image_name:
+        full_image_name = '{}:latest'.format(image_name)
     for image in docker_client.images():
-        if any(image_name in tag for tag in image['RepoTags']):
+        if full_image_name in image['RepoTags']:
             break
     else:
         split = image_name.split(':')
```
jsingle: fix the line below this that uses latest too right?

thieman: The `else` is fine as far as I can tell, the problem was we were falsely believing we had the image we needed checked out when we didn't.

jsingle:
Seems to make sense

PR #518, opened 2015-08-18
errors in test image creation are caught

### LINKED ISSUE #516: Tests should fail immediately if any command in `test.once` fails
Currently, a step specified in the `once` section of a test spec can fail silently.  These commands should be run with `set -e`

jsingle:
@thieman 

Resolves #516


dusty/commands/test.py
```diff
@@ -70,7 +70,11 @@ def run_app_or_lib_tests(app_or_lib_name, suite_name, test_arguments, should_exi
     expanded_specs = get_expanded_libs_specs()
     spec = expanded_specs.get_app_or_lib(app_or_lib_name)
     test_command = _construct_test_command(spec, suite_name, test_arguments)
-    ensure_test_image(client, app_or_lib_name, expanded_specs, force_recreate=force_recreate)
+    try:
+        ensure_test_image(client, app_or_lib_name, expanded_specs, force_recreate=force_recreate)
+    except ImageCreationError as e:
+        log_to_client('Failed to create test container with error {}'.format(e.code))
+        sys.exit(e.code)
```
thieman: Is anything gonna show me the actual error?

tests/fixtures.py
```diff
@@ -126,19 +126,31 @@ def specs_fixture_with_depends():
                             }})
     _write('service', 'servica', {'image': 'busybox'})
 
-def fixture_with_commands(once_fail=False, always_fail=False):
-    once_commands = ['touch /once_test_file', 'echo "once ran" | tee -a /once_test_file']
+def fixture_with_commands(once_fail=False, always_fail=False, test_fail=False):
+    once_commands = ['touch /once_test_file', 'stdbuf -i0 -o0 -e0 echo "once ran" | tee -a /once_test_file']
```
thieman: lol wat

dusty/commands/test.py
jsingle: It would be printed right above 'Failed to create test container with error 127'

tests/fixtures.py
jsingle: :)  Flushes stuff through the pipe, since it's getting teed and redirected and fancy stuff

jsingle: http://stackoverflow.com/questions/11337041/force-line-buffering-of-stdout-when-piping-to-tee

tests/testcases.py
```diff
@@ -156,9 +159,13 @@ def run_command(self, args, fake_exit, raise_on_error=True):
...
         stdout generated by the command."""
         with patch('dusty.commands.utils.exec_docker', wraps=self.exec_docker_patch) as fake_exec_docker:
+            fake_exit.side_effect = SysExit('sys.exit called')
             self.fake_exec_docker = fake_exec_docker
             sys.argv = ['dusty'] + args.split(' ')
-            client_entrypoint()
+            try:
+                client_entrypoint()
+            except SysExit:
+                pass
```
thieman: Is there a better way to do what you're trying to do here?

tests/integration/command_file_test.py
```diff
@@ -12,7 +12,10 @@ def tearDown(self):
             self.run_command('stop')
         except:
             pass
-        self.remove_container('appa')
+        try:
+            self.remove_container('appa')
+        except RuntimeError:
```
thieman: You can be general here, it's a tearDown

tests/fixtures.py
thieman: woah

tests/testcases.py
jsingle: I need sys.exit to actually "return" (stop executing), but not actually call sys.exit.  I think raising and then catching an exception is the only way to do that...

PR #523, opened 2015-08-19
status handles powered of VM gracefully

### LINKED ISSUE #521: `dusty status` should fail gracefully when boot2docker/docker-machine isn't started
Currently `dusty status` seems to make no attempt to check if the VM is started.  It also does not print a nice error message

jsingle:
@thieman 

Resolves #521


thieman:
LGTM "reslove"

jsingle:
seplling how does it work

PR #527, opened 2015-08-19
Js bundle services

### LINKED ISSUE #465: Bundles should be able to specify services
I think this would mainly be useful for allowing optional bundles which start up services which can then be specified through `conditional_links` to applications, essentially allowing optional services.

jsingle:
@thieman 

Resolves #465


docs/specs/bundle-specs.md
```diff
@@ -28,3 +28,14 @@ apps:
...
 your app and service specs.
+
+## services
+
+```
+services:
+  - postgres
+```
+
+An optional list of services which will be run by `dusty up` if this bundle
```
thieman: The use of the word "optional" here makes it seem like apps are mandatory. The actual constraint is that at least one app or service must be provided, right?

jsingle: You're right, that should be clarified

docs/changelog.md
```diff
@@ -8,6 +8,7 @@
     * The `restart` command no longer takes a `--no-sync` flag, due to the move to NFS.
 * **New**
     * Dusty now attempts to fix a known problem with boot2docker networking, resulting in up to a 10x improvement in network performance. See [this boot2docker issue](https://github.com/boot2docker/boot2docker/issues/1022) for more information.
+    * Dusty bundles can now specify services, just as they specify apps
```
thieman: Don't need "Dusty" here

dusty/compiler/spec_assembler.py
```diff
@@ -67,12 +67,16 @@ def _get_referenced_libs(specs):
...
+    Returns all services that are referenced in specs.apps.depends.services,
+    or in specs.bundles.services
     """
     active_services = set()
     for app_spec in specs['apps'].values():
         for service in app_spec['depends']['services']:
             active_services.add(service)
+    for bundle_spec in specs['bundles'].values():
+        for service in bundle_spec['services']:
+            active_services.add(service)
```
thieman: That was easy

jsingle: :smile:

thieman:
LGTM

PR #535, opened 2015-08-20
add functionality to wait for docker exec processes

jsingle:
@thieman I feel like if we're setting timeouts, we should let the calling test say what the timeout should be

### DIFF
--- a/tests/integration/cli/scripts_test.py
+++ b/tests/integration/cli/scripts_test.py
@@ -33,13 +33,16 @@ class TestScriptsCLI(DustyIntegrationTestCase):
     def test_basic(self):
         self.assertFileNotInContainer('appa', '/app/a/foo')
         self.run_command('scripts appa example')
+        self.wait_for_exec_docker(timeout=1)
         self.assertFileInContainer('appa', '/app/a/foo')
         self.exec_in_container('appa', 'rm /app/a/foo')
 
     def test_with_arg(self):
         self.run_command('scripts appa example')
+        self.wait_for_exec_docker(timeout=1)
         self.assertFileInContainer('appa', '/app/a/foo')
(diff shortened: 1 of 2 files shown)


tests/integration/cli/scripts_test.py
```diff
@@ -33,13 +33,16 @@ def tearDown(self):
     def test_basic(self):
         self.assertFileNotInContainer('appa', '/app/a/foo')
         self.run_command('scripts appa example')
+        self.wait_for_exec_docker(1)
```
thieman: I think it makes sense to provide a reasonable default, something like 3 seconds. If a test is going longer than that it's probably a crappy test, hopefully seeing someone override the default timeout would tip off a reviewer that there's a problem.

tests/testcases.py
```diff
@@ -177,6 +176,17 @@ def run_command(self, args, fake_exit, raise_on_error=True):
             self._clear_stdout()
             return result
 
+    def wait_for_exec_docker(self, timeout):
+        def wait_for_processes():
+            for p in self.exec_docker_processes:
+                if p.returncode is None:
+                    p.communicate()
```
thieman: Does the `returncode` get set even if you haven't called any follow-up functions like `communicate` on your `Popen` object?

tests/integration/cli/scripts_test.py
jsingle: sounds good

thieman:
LGTM

tests/testcases.py
jsingle: Hmm I should really be using `poll` here

jsingle:
Well this is working locally, but it sure isn't in jankins :rage:

PR #541, opened 2015-08-24
Parallelize Git operations

### LINKED ISSUE #536: Parallelize Git pulls over managed repos
Compose's recent improvement to parallelize stopping and removing containers have made some Dusty commands way faster. The repo updating we do is still serial and embarrassing when compared to what Compose is doing. Can we parallelize the Git operations on managed repos to make `dusty up` and similar commands much faster?

thieman:
@jsingle Closes #536. Maybe worth waiting until `0.6.0` for this one so we can get some RC testing done on it.

### DIFF
--- a/dusty/parallel.py
+++ b/dusty/parallel.py
@@ -1,0 +1,55 @@
+"""Utilities for multithreaded parallel execution of tasks."""
+
+import threading
+import multiprocessing
+from contextlib import contextmanager
+from Queue import Queue
+
+from .log import log_to_client
+
+class TaskQueue(Queue, object):
+    """Executable task queue used for multithreaded execution of multiple
+    function calls. Concurrency is limited by `pool_size`."""
+    def __init__(self, pool_size):
+        super(TaskQueue, self).__init__()
+        self.threads = []
+        self.errors = []
+        self.pool_semaphore = threading.Semaphore(pool_size)
+
+    def enqueue_task(self, fn, *args, **kwargs):
+        self.put((fn, args, kwargs))
+
+    def _task_executor(self, fn, args, kwargs):
+        try:
+            fn(*args, **kwargs)
+        except Exception as e:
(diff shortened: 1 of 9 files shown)


dusty/parallel.py
```diff
@@ -0,0 +1,47 @@
...
+        try:
+            fn(*args, **kwargs)
+        except Exception as e:
+            self.errors.append(e.message)
+
+    def execute(self):
+        while not self.empty():
+            fn, args, kwargs = self.get()
+            self.pool.apply_async(self._task_executor, args=(fn, args, kwargs))
+        self.pool.close()
```
inlinestyle: Since it looks like you can't reuse the pool after you've closed it, I think we should do one of the following:
1. Do something to indicate that TaskQueue should not invoked directly outside of parallel.py
2. Create the pool inside `execute`

inlinestyle:
:space_invader: :+1:

dusty/parallel.py
```diff
@@ -0,0 +1,48 @@
...
+        self.pool.join()
+
+        if self.errors:
+            for error in self.errors:
+                log_to_client(error)
+            raise RuntimeError("Exceptions encountered during parallel task execution")
+
+@contextmanager
+def parallel_task_queue(pool_size=multiprocessing.cpu_count()*2):
```
jsingle: Curious as to why you chose this exact pool size

thieman: I think this is a fairly standard go-to for IO-bound parallelism. </handwave>

inlinestyle: I think it's # of hyperthreaded cores?

thieman: Yeah that func returns 8 on my MBP

jsingle: Cool

jsingle:
lgtm

PR #544, opened 2015-08-25
Js nginx name conflict

### LINKED ISSUE #530: Can't name apps or services `nginx`
This conflicts with naming of dusty's nginx for internal use, basically clobbering it.  Discovered by @melkorm 

The fix here should probably just be to add validation preventing apps and services from being named nginx.  Currently it's way to hard to tell what's actually going on

jsingle:
@zacharym 

Resolves #530


zacharym:
lgtm

PR #546, opened 2015-08-25
dusty env command

### LINKED ISSUE #474: RFC: `dusty env` to allow user to specify environment overrides
It can often be useful to change the behavior of an application by manipulating environment variables rather than code. A common example of this would be changing the log level to a more verbose setting in order to debug an issue. Currently, the only ways to do this with Dusty are to override the setting in the code (which doesn't work for services) or override the Dusty specs and override the environment in the `compose` key (which is clumsy and doesn't scale).

I'm proposing the addition of an `env` CLI subcommand which would allow the user to specify environment variable overrides to containers (apps or services) run by Dusty. These overrides would be specific to the user's machine and would not pollute the Dusty specs. Something like:

```
dusty env set <app or service name> <variable name> <variable value>
# e.g.
dusty env set website LOG_LEVEL debug
```

During `dusty up` and similar commands, Dusty would merge the current overrides into any existing environment settings in the specs, with preference being given to the overrides.

We would also need commands to unset the overrides and display to the user which environment overrides are currently registered, e.g.

```
dusty env unset <app or service name> <variable name>
dusty env list <app or service name>
```

jsingle:
@thieman still gotta write tests but it works.  The main code is ready for you to start looking at

Resolves #474

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -7,6 +7,7 @@
     * The setup override flag `--boot2docker_vm_memory` has been renamed to `--vm_memory`
 * **New**
     * Managed repos are now updated in parallel during `dusty up` and `dusty test`.
+    * New command `dusty env` can set environment variable overrides per app or service. These variables can also be read from a file, as with Docker Compose's `env_file` key.
 * **Misc**
     * Dusty now uses the name `dustyInternalNginx` for its own nginx container, to make naming conflicts less likely.
 
--- a/dusty/cli/env.py
+++ b/dusty/cli/env.py
@@ -1,0 +1,40 @@
+"""Set environment variable overrides.
+
+Environment variables specified will be added to app
+and service container environments, overriding variables
+specified in a `compose.env` spec (if present).
+
+Usage:
+  env list [<app_or_service>]
+  env set <app_or_service> (<var_name> <value> | --file <local_file>)
+  env unset <app_or_service> (--all | <var_name>)
+
+Commands:
+  list        List all environment variables and their current values.
+  set         Set a variable name to a new value for the given app or service.
+  unset       Unset a variable for the given app or service.
+"""
+import os
+
+from docopt import docopt
+
+from ..payload import Payload
+from ..commands import env
+
+def main(argv):
+    args = docopt(__doc__, argv)
+    if args['list']:
+        if args['<app_or_service>']:
+            return Payload(env.list_app_or_service, args['<app_or_service>'])
+        else:
+            return Payload(env.list_all)
+    elif args['set']:
+        if args['--file']:
+            return Payload(env.set_from_file, args['<app_or_service>'], os.path.abspath(args['<local_file>']))
+        else:
+            return Payload(env.set_var, args['<app_or_service>'], args['<var_name>'], args['<value>'])
+    elif args['unset']:
+        if args['--all']:
+            return Payload(env.unset_all, args['<app_or_service>'])
+        else:
+            return Payload(env.unset_var, args['<app_or_service>'], args['<var_name>'])
--- a/dusty/commands/env.py
+++ b/dusty/commands/env.py
@@ -1,0 +1,107 @@
+import os
+
+from prettytable import PrettyTable
+
+from ..log import log_to_client
+from ..compiler.spec_assembler import get_assembled_specs, get_specs
+from ..config import get_env_config, save_config_value
+from .. import constants
+from ..payload import daemon_command
+
+def _save_env(env_config):
+    save_config_value(constants.CONFIG_ENV_KEY, env_config)
+
+def _assert_app_or_service_exists(app_or_service):
+    """
+    get_app_or_service will throw a KeyError if the app_or_service
+    is not found.
+    """
+    get_specs().get_app_or_service(app_or_service)
+
+@daemon_command
+def list_app_or_service(app_or_service):
+    _assert_app_or_service_exists(app_or_service)
+    env_dict = get_env_config().get(app_or_service)
+    if not env_dict:
+        log_to_client('No environment overrides specified for {}'.format(app_or_service))
+        return
+    table = PrettyTable(["Variable", "Value"])
+    for var_name, value in env_dict.iteritems():
+        table.add_row([var_name, value])
+    log_to_client(table.get_string())
+
+@daemon_command
+def list_all():
+    env_config = get_env_config()
+    if not env_config:
+        log_to_client('No environment overrides specified')
+        return
+    active_containers = set([cont.name for cont in get_assembled_specs().get_apps_and_services()])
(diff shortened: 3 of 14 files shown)


dusty/commands/env.py
```diff
@@ -0,0 +1,94 @@
+import functools
```
jsingle: oops UNUSED

dusty/commands/env.py
```diff
@@ -0,0 +1,94 @@
...
+from prettytable import PrettyTable
+
+from ..log import log_to_client
+from ..compiler.spec_assembler import get_assembled_specs, get_specs
+from ..config import get_env_config, save_config_value
+from .. import constants
+from ..payload import daemon_command
+
+def _save_env(global_env_dict):
+    save_config_value(constants.CONFIG_ENV_KEY, global_env_dict)
```
thieman: Never use `global` in a variable name unless it's actually a global variable, which this isn't

dusty/commands/env.py
```diff
@@ -0,0 +1,94 @@
...
+    for app_or_service, env_dict in global_env_dict.iteritems():
+        if app_or_service in active_containers:
+            for var_name, value in env_dict.iteritems():
+                table.add_row([app_or_service, var_name, value])
+    log_to_client(table.get_string())
+
+@daemon_command
+def set_from_file(app_or_service, local_file_path):
+    _assert_app_or_service_exists(app_or_service)
+    if not os.path.exists(local_file_path):
```
thieman: This should also check whether the file is readable, I think you can use `os.stat` for that

dusty/commands/env.py
jsingle: haha good point

dusty/commands/env.py
```diff
@@ -0,0 +1,94 @@
...
+    if not global_env_dict.get(app_or_service):
+        global_env_dict[app_or_service] = {}
+    with open(local_file_path, 'r') as f:
+        for line in f.readlines():
+            equal_split = line.split('=', 1)
+            if len(equal_split) != 2:
+                raise RuntimeError('Line {} is not of the format VARIABLE=value'.format(line))
+            var_name = equal_split[0]
+            value = equal_split[1].rstrip()
+            global_env_dict[app_or_service][var_name] = value
```
thieman: Find a parser for this. Env files support comments and stuff.

dusty/commands/env.py
jsingle: K

tests/integration/cli/env_test.py
```diff
@@ -0,0 +1,58 @@
...
+    def setUp(self):
+        super(TestEnvCLI, self).setUp()
+        single_specs_fixture()
+        local_dir = Repo('github.com/app/a').managed_path
+        if not os.path.exists(local_dir):
+            os.makedirs(local_dir)
+        self.run_command('bundles activate bundle-a')
+
+    def tearDown(self):
+        self.run_command('stop')
```
thieman: Don't we usually put this in a `try`? Can't remember exactly why

tests/integration/cli/env_test.py
```diff
@@ -0,0 +1,58 @@
...
+        self.run_command('env unset appa --all')
+        super(TestEnvCLI, self).tearDown()
+
+    def assertEnvInContainer(self, app_or_service, var, value):
+        env = self.exec_in_container(app_or_service, 'printenv {}'.format(var)).rstrip()
+        self.assertEqual(env, value)
+
+    def assertEnvNotInContainer(self, app_or_service, var):
+        env = self.exec_in_container(app_or_service, 'printenv {}'.format(var)).rstrip()
+        self.assertEqual(env, '')
```
thieman: These should live in the testcase with the other asserts

tests/integration/cli/env_test.py
```diff
@@ -0,0 +1,58 @@
...
+        self.run_command('env set appa pitcher pedro')
+        self.run_command('up --no-pull')
+        self.assertEnvInContainer('appa', 'pitcher', 'pedro')
+
+    def test_env_overrides_spec_environment(self):
+        self.run_command('up --no-pull')
+        self.assertEnvInContainer('appa', 'SPEC_VALUE', 'spec-specified-value')
+        self.run_command('env set appa SPEC_VALUE new-value')
+        self.run_command('up --no-pull')
+        self.assertEnvInContainer('appa', 'SPEC_VALUE', 'new-value')
```
thieman: Can you rewrite this to only need to do one up, instead specifying two different env vars and only overwriting one?

tests/integration/cli/env_test.py
```diff
@@ -0,0 +1,58 @@
...
+        self.assertEnvInContainer('appa', 'SPEC_VALUE', 'spec-specified-value')
+
+    def test_unset_one(self):
+        self.run_command('env set appa pitcher pedro')
+        self.run_command('env set appa SPEC_VALUE new-value')
+        self.run_command('up --no-pull')
+        self.run_command('env unset appa SPEC_VALUE')
+        self.run_command('up --no-pull')
+        self.assertEnvInContainer('appa', 'pitcher', 'pedro')
+        self.assertEnvInContainer('appa', 'SPEC_VALUE', 'spec-specified-value')
```
thieman: Can these two also be made to only do `up` once?

dusty/cli/env.py
```diff
@@ -0,0 +1,40 @@
...
+Usage:
+  env list [<app_or_service>]
+  env set <app_or_service> (<var_name> <value> | --file <local_file>)
+  env unset <app_or_service> (--all | <var_name>)
+
+Commands:
+  list        List all environment variables and their current values.
+  set         Set a variable name to a new value for the given app or service.
+  unset       Unset a variable for the given app or service.
+"""
```
thieman: Throw this in the usage docs. Also add this feature to changelog.

tests/integration/cli/env_test.py
jsingle: hmm we often do.  I'm also unsure why

dusty/cli/env.py
jsingle: Done

jsingle:
I'm thinking that integration test failure is not related to this.  Mac slave might be sad b/c zach pushed a non-rebased PR

docs/changelog.md
```diff
@@ -7,6 +7,7 @@
     * The setup override flag `--boot2docker_vm_memory` has been renamed to `--vm_memory`
 * **New**
     * Managed repos are now updated in parallel during `dusty up` and `dusty test`.
+    * New command `dusty env` can set environment variable overrides.
```
thieman: Flesh this out a little

thieman:
LGTM, please squash

PR #548, opened 2015-08-26
Zm unicode out

zacharym:
@thieman it might have been this simple

### DIFF
--- a/dusty/payload.py
+++ b/dusty/payload.py
@@ -3,6 +3,14 @@ import yaml
 
 from .constants import VERSION
 
+def _encode_utf8(args):
+    for idx, arg in enumerate(args):
+        if isinstance(arg, basestring):
+            args[idx] = arg.encode('utf-8')
+        if isinstance(arg, list):
+            args[idx] = _encode_utf8(arg)
+    return args
+
 class Payload(object):
     def __init__(self, fn, *args, **kwargs):
         self.fn = fn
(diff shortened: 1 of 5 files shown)


zacharym:
@thieman I re-ran these puppies from earlier to show the tests work. Going to go through and add unicodes around the other fixtures

thieman:
Cool

zacharym:
@jsingle I think this takes care of us for encoding arguments, we still have inconsistency given how we are reading specs. cc @thieman

dusty/payload.py
```diff
@@ -30,7 +38,9 @@ def serialize(self):
 
     @staticmethod
     def deserialize(doc):
-        return yaml.safe_load(doc)
+        loaded = yaml.safe_load(doc)
+        loaded['args'] = _encode_utf8(loaded['args'])
```
jsingle: what about kwargs?

dusty/payload.py
```diff
@@ -3,6 +3,14 @@
 
 from .constants import VERSION
 
+def _encode_utf8(args):
```
jsingle: I'm concerned about having to traverse a data structure like this.  We should talk about it

zacharym: I'll look into the root cause of the issues with `yaml.safe_load` and see if there is something we can do on that side

PR #549, opened 2015-08-26
provide strs to execv

zacharym:
@jsingle this was breaking for me locally .. not sure if I'm doing something else stupidly locally
cc @thieman


jsingle:
Certainly won't hurt anything, I'm definitely curious as to why you were seeing this (but not others)

thieman:
-1 fix the actual problem

jsingle:
Yeah there's an inconsistency in the different ways we're setting the `vm_memory_size` field.  The fix isn't just to cast to a string here

zacharym:
Yeah, this was more just a convenient way to share information, not what I thought the actual fix would be.

PR #552, opened 2015-08-26
save vm_memory as a str

zacharym:
@jsingle I think this takes care of it. Do you know why we would have this cast as an int originally?


jsingle:
I think that we had it as an int because it's conceptually an int, not a string

jsingle:
I think this is fine though

thieman:
:bear:

PR #553, opened 2015-08-26
New dusty assets command

### LINKED ISSUE #470: RFC: `dusty assets` to get user keys/secrets into containers
Original request from user `abooth` in the Slack org. Currently, it is very difficult to create a Dusty app which relies on a user-provided secret, for example a private SSH key. The options are limited: you can bake the secret into the image (if and only if you're using a private Docker registry), or you can have the users manually `dusty cp` their secret into the container once it's created. Both of these leave a lot to be desired.

This is probably a fairly common thing to want to do, as you could create an app which installs packages from a private package manager or private GitHub repo. Dusty should support this flow in a way that doesn't require private Docker registries or manual user intervention.

My proposal is to add a concept called an `asset` which would allow an app or library spec to specify an asset which should be provided by the user and made available at a specific path inside the container. We may also need to add these spec options to the testing containers.

```yaml
# inside an app or lib spec
assets:
  - name: GITHUB_KEY
    path: /etc/keys/id_github.rsa
```

In addition to this spec change, we would add an `assets` subcommand to the CLI. This would allow users to register assets at a local filepath on their Mac. Something like:

```
dusty assets register GITHUB_KEY ~/.ssh/my_github_key.rsa
```

We would need standard lifecycle management around these assets.

Finally, `dusty up` should check the required assets for the apps and libs that are active and throw an error (before trying to start anything) if any required assets have not been registered.

The details of keeping the assets up-to-date inside the VM may be drastically changed by #442, which is currently in progress. It may make sense to hold off on implementing this until NFS is completed, but we can talk out the design now.

jsingle:
@thieman seems to be working, but tests will need a bunch of work.  This code is ready for input though cc @zacharym 

Resolves #470

### DIFF
--- a/dusty/cli/__init__.py
+++ b/dusty/cli/__init__.py
@@ -4,6 +4,7 @@ Usage:
   dusty [options] [<command>] [<args>...]
 
 Commands:
+  assets     Manage files that don't live in version control
   bundles    Manage which sets of applications to run
   config     Get or set Dusty config variables
   cp         Copy files between local filesystem and containers
@@ -43,11 +44,12 @@ from ..daemon import main as run_daemon
 from ..config import get_config_value
 from ..log import configure_client_logging, log_to_client
 from ..payload import Payload
-from . import (bundles, config, cp, dump, disk, env, logs, repos, restart, scripts, shell, stop,
+from . import (assets, bundles, config, cp, dump, disk, env, logs, repos, restart, scripts, shell, stop,
                up, upgrade, validate, version, setup, test, status)
 from .. import constants
 
 MODULE_MAP = {
+    'assets': assets,
     'bundles': bundles,
     'config': config,
     'cp': cp,
--- a/dusty/cli/assets.py
+++ b/dusty/cli/assets.py
@@ -1,0 +1,49 @@
+"""
+Place files in Dusty containers
+
+Assets are files to be put in containers, but which don't live in a repository.
+Assets are declared in Dusty specs of apps and libraries, and their values are
+managed with the CLI.
+
+Usage:
+    assets list [<app_or_lib>]
+    assets read <asset_key>
+    assets set <asset_key> <local_path>
+    assets unset <asset_key>
+
+Commands:
+    list        List all assets that are defined in specs for active apps and libs
+    read        Print the current value of an asset
+    set         Associate an asset with the contents of a local file
+    unset       Delete the currently registered value of an asset
+
+Examples:
+    To set the value of the asset GITHUB_KEY to the contents of ~/.ssh/id_rsa:
+        dusty assets set GITHUB_KEY ~/.ssh/id_rsa
+"""
+
+import os
+import sys
+
+from docopt import docopt
+
+from ..payload import Payload
+from ..commands import assets
+from ..log import log_to_client
+
+def main(argv):
+    args = docopt(__doc__, argv)
+    if args['list']:
+        if args['<app_or_lib>']:
+            assets.list_by_app_or_lib(args['<app_or_lib>'])
+        else:
+            assets.list_all()
+    elif args['read']:
+        assets.read_asset(args['<asset_key>'])
+    elif args['set']:
+        if not os.access(args['<local_path>'], os.R_OK):
+            log_to_client('Local path {} does not exist, or you don\'t have permission to access it'.format(args['<local_path>']))
+            sys.exit(1)
+        assets.set_asset(args['<asset_key>'], os.path.abspath(args['<local_path>']))
+    elif args['unset']:
+        assets.unset_asset(args['<asset_key>'])
--- a/dusty/commands/assets.py
+++ b/dusty/commands/assets.py
@@ -1,0 +1,42 @@
+import os
+
+from prettytable import PrettyTable
+
+from ..compiler.spec_assembler import get_specs, get_assembled_specs
+from ..systems.virtualbox import asset_is_set, asset_value, asset_vm_path, remove_asset
+from ..systems.rsync import sync_local_path_to_vm
+from ..log import log_to_client
+
+def list_by_app_or_lib(app_or_lib):
+    spec = get_specs().get_app_or_lib(app_or_lib)
+    table = PrettyTable(["Asset", "Is Set", "Required", "In-Container Path"])
+    for asset in spec['assets']:
+        table.add_row([asset['name'], 'X' if asset_is_set(asset['name']) else '', 'X' if asset['required'] else '', asset['path']])
+    log_to_client(table.get_string())
+
+def _get_string_of_set(items):
+    return ', '.join(sorted(items))
+
+def list_all():
+    table = PrettyTable(["Asset", "Is Set", "Used By", "Required By"])
+    assembled_specs = get_assembled_specs()
+    for asset_name, asset_info in assembled_specs['assets'].iteritems():
+        used_by_display = _get_string_of_set(asset_info['used_by'])
+        required_by_display = _get_string_of_set(asset_info['required_by'])
+        table.add_row([asset_name, 'X' if asset_is_set(asset_name) else '', used_by_display, required_by_display])
+    log_to_client(table.get_string())
+
+def read_asset(asset_key):
+    if not asset_is_set(asset_key):
+        log_to_client('Asset {} isn\'t set'.format(asset_key))
+        return
+    log_to_client(asset_value(asset_key))
+
+def set_asset(asset_key, local_path):
+    sync_local_path_to_vm(local_path, asset_vm_path(asset_key), demote=True)
+
+def unset_asset(asset_key):
+    if not asset_is_set(asset_key):
+        log_to_client('Asset {} isn\'t set'.format(asset_key))
+        return
+    remove_asset(asset_key)
--- a/dusty/compiler/spec_assembler.py
+++ b/dusty/compiler/spec_assembler.py
@@ -93,6 +93,22 @@ def _filter_active(spec_type, specs):
             del specs[spec_type][item_name]
     logging.info("Spec Assembler: filtered active {} to {}".format(spec_type, set(specs[spec_type].keys())))
 
+def _add_active_assets(specs):
+    """
+    This function adds an assets key to the specs, which is filled in with a dictionary
+    of all assets defined by apps and libs in the specs
+    """
+    specs['assets'] = {}
+    for spec in specs.get_apps_and_libs():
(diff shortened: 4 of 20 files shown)


dusty/compiler/spec_assembler.py
```diff
@@ -93,6 +93,18 @@ def _filter_active(spec_type, specs):
...
+    specs['assets'] = {}
+    for spec in specs.get_apps_and_libs():
+        for asset in spec['assets']:
+            if not specs['assets'].get(asset['name']):
+                specs['assets'][asset['name']] = {}
+                specs['assets'][asset['name']]['required_by'] = set()
+                specs['assets'][asset['name']]['used_by'] = set()
+            specs['assets'][asset['name']]['used_by'].add(spec.name)
+            if asset['required']:
+                specs['assets'][asset['name']]['required_by'].add(spec.name)
```
jsingle: I'm wondering if instead of this function, I should expand assets inside apps specs.  So that an app spec will have all the assets that it and all of its libs need in the same place

dusty/cli/__init__.py
```diff
@@ -4,6 +4,7 @@
   dusty [options] [<command>] [<args>...]
 
 Commands:
+  assets     Set assets to place in containers
```
thieman: Take another shot at this sentence

dusty/cli/assets.py
```diff
@@ -0,0 +1,44 @@
+"""
+Place special files in Dusty containers
+
+Assets are special files on your local filesystem, not located in
+repositories, that Dusty will put in containers for you.
```
thieman: Let's try to clean up the terminology here

dusty/cli/assets.py
```diff
@@ -0,0 +1,44 @@
...
+    assets list [<app_or_lib>]
+    assets read <asset_key>
+    assets set <asset_key> <local_path>
+    assets unset <asset_key>
+
+Commands:
+    list        List all assets that are defined in specs for active apps and libs
+    read        Print the current value of an asset
+    set         Associate an asset with the contents of a local file
+    unset       Delete the currently registered value of an asset
```
thieman: Can we add some examples?

jsingle: Do examples belong in these docstrings?

dusty/commands/assets.py
```diff
@@ -0,0 +1,42 @@
...
+
+def list_by_app_or_lib(app_or_lib):
+    spec = get_specs().get_app_or_lib(app_or_lib)
+    table = PrettyTable(["Asset", "Is Set", "Required", "In-Container Path"])
+    for asset in spec['assets']:
+        table.add_row([asset['name'], 'X' if asset_is_set(asset['name']) else '', 'X' if asset['required'] else '', asset['path']])
+    log_to_client(table.get_string())
+
+def _get_set_display(items):
+    return ', '.join(sorted(items))
```
thieman: Doesn't really seem worth a function

dusty/systems/virtualbox/__init__.py
```diff
@@ -131,3 +141,29 @@ def get_host_ip():
...
+    for asset_key, asset_info in assembled_specs['assets'].iteritems():
+        if asset_info['required_by'] and not asset_is_set(asset_key):
+            required_and_absent.append(asset_key)
+    return required_and_absent
+
+def asset_vm_path(asset_key):
+    return os.path.join(constants.VM_ASSETS_DIR, asset_key)
+
+def asset_is_set(asset_key):
+    return _file_exists_on_vm(asset_vm_path(asset_key))
```
thieman: I think we probably want to batch this. Doing a bunch of individual SSH commands will get slow.

dusty/commands/assets.py
```diff
@@ -0,0 +1,42 @@
...
+        table.add_row([asset_name, 'X' if asset_is_set(asset_name) else '', used_by_display, required_by_display])
+    log_to_client(table.get_string())
+
+def read(asset_key):
+    if not asset_is_set(asset_key):
+        log_to_client('Asset {} isn\'t set'.format(asset_key))
+        return
+    log_to_client(asset_value(asset_key))
+
+def set_asset(asset_key, local_path):
```
thieman: Try to standardize your naming conventions, if this one is `set_asset` then the one above should be `read_asset`. I think since these aren't inside a class it's good to have the noun be part of the function name.

dusty/systems/virtualbox/__init__.py
jsingle: that's true.  Seems complicated to batch tho

dusty/systems/virtualbox/__init__.py
```diff
@@ -13,10 +13,13 @@
...
 
-def _run_command_on_vm(command_list):
+def _run_command_on_vm(command_list, return_output=False):
     ssh_command = ['docker-machine', 'ssh', constants.VM_MACHINE_NAME]
     ssh_command.extend([command_list])
-    check_and_log_output_and_error_demoted(ssh_command)
+    if return_output:
+        return check_output_demoted(ssh_command)
+    else:
+        check_and_log_output_and_error_demoted(ssh_command)
```
thieman: What are you actually trying to toggle on here, returning the output or logging the output to the client? This is currently doing both (complex!).

dusty/compiler/spec_assembler.py
thieman: Can you walk me through how assets are being used in the specs? Probably worth throwing some docstrings on this as well.

dusty/systems/virtualbox/__init__.py
jsingle: Yeah the idea is to toggle both.  The function structure could probably be improved

dusty/systems/virtualbox/__init__.py
```diff
@@ -33,11 +36,17 @@ def _ensure_persist_dir_is_linked():
...
+def _ensure_vm_dir_exists(vm_dir):
+    logging.info('Creating {} in VM to support dusty'.format(vm_dir))
+    mkdir_if_cmd = 'if [ ! -d {0} ]; then sudo mkdir {0}; fi'.format(vm_dir)
     _run_command_on_vm(mkdir_if_cmd)
 
+def _ensure_cp_dir_exists():
+    _ensure_vm_dir_exists(constants.VM_CP_DIR)
+
+def _ensure_assets_dir_exists():
+    _ensure_vm_dir_exists(constants.VM_ASSETS_DIR)
```
thieman: Cool refactor bro

tests/integration/cli/assets_test.py
```diff
@@ -0,0 +1,80 @@
...
+            f.write('required_app_contents')
+        self.optional_app_file = tempfile.mkstemp()[1]
+        with open(self.optional_app_file, 'w') as f:
+            f.write('optional_app_contents')
+        self.required_lib_file = tempfile.mkstemp()[1]
+        with open(self.required_lib_file, 'w') as f:
+            f.write('required_lib_contents')
+        self.optional_lib_file = tempfile.mkstemp()[1]
+        with open(self.optional_lib_file, 'w') as f:
+            f.write('optional_lib_contents')
```
thieman: RYE

tests/integration/cli/assets_test.py
```diff
@@ -0,0 +1,80 @@
...
+from dusty.source import Repo
+from ...testcases import DustyIntegrationTestCase
+from ...fixtures import assets_fixture
+
+class TestAssetsCLI(DustyIntegrationTestCase):
+    def setUp(self):
+        super(TestAssetsCLI, self).setUp()
+        assets_fixture()
+
+        local_dir = Repo('github.com/lib/a').managed_path
```
thieman: How do we normally handle this in integration tests? I thought we had a way of throwing these in tmpdirs so they didn't clutter the real /etc/dusty

tests/integration/cli/assets_test.py
```diff
@@ -0,0 +1,80 @@
...
+        except:
+            pass
+
+    def test_asset_in_container_app(self):
+        self.run_command('up --no-pull')
+        self.assertFileContentsInContainer('app-a', '/required_app_path', 'required_app_contents')
+
+    def test_asset_in_container_lib(self):
+        self.run_command('up --no-pull')
+        self.assertFileContentsInContainer('app-a', '/required_lib_path', 'required_lib_contents')
```
thieman: Motion to consolidate these two tests to eliminate an `up`

tests/integration/cli/assets_test.py
jsingle: Hmm we might just override them to a tmp location

dusty/cli/assets.py
thieman: Also remember usage docs and changelog

thieman: And specs docs! Wooo

tests/integration/cli/assets_test.py
jsingle: Sure

dusty/commands/assets.py
jsingle: Yeah I did this originally because `set` is a special python thing lol.  I got really confused for a few minutes lol

dusty/systems/virtualbox/__init__.py
thieman: ༼ つ ◕_◕ ༽つ THE POWER OF LS ༼ つ ◕_◕ ༽つ

dusty/systems/virtualbox/__init__.py
```diff
@@ -131,3 +141,29 @@ def get_host_ip():
             if line.startswith('IPAddress'):
                 return line.split()[1]
     raise RuntimeError('Host IP in host only network not found')
+
+def _file_exists_on_vm(file_path):
+    result = _run_command_on_vm('if [ -f {} ]; then echo yes; else echo no; fi'.format(file_path),
+                                return_output=True).rstrip()
+    if result == 'yes':
+        return True
+    return False
```
thieman: `return result == 'yes'`

dusty/systems/virtualbox/__init__.py
thieman: Is there any reason to not always return? What's the value of returning None all the time?

jsingle: nothing.  I suppose the real difference is either logging to client or not logging to client

dusty/systems/virtualbox/__init__.py
jsingle: smrt

PR #560, opened 2015-08-28
small refactor of rsync logging

jsingle:
@paetling not sure why we were doing this before

### DIFF
--- a/dusty/commands/assets.py
+++ b/dusty/commands/assets.py
@@ -33,7 +33,7 @@ def read_asset(asset_key):
     log_to_client(asset_value(asset_key))
 
 def set_asset(asset_key, local_path):
-    sync_local_path_to_vm(local_path, asset_vm_path(asset_key), demote=True)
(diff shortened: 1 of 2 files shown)


dusty/commands/assets.py
```diff
@@ -33,7 +33,7 @@ def read_asset(asset_key):
     log_to_client(asset_value(asset_key))
 
 def set_asset(asset_key, local_path):
-    sync_local_path_to_vm(local_path, asset_vm_path(asset_key), demote=True)
+    sync_local_path_to_vm(local_path, asset_vm_path(asset_key))
```
paetling: its not a permissions thing is it?  probably should not be because you run as root in the container.

paetling:
seems fine i guess. Not really sure why it was like this to begin with?

dusty/commands/assets.py
jsingle: This particular change doesn't really matter.  Basically just a cleanup of my earlier testing

PR #562, opened 2015-08-31
Js docker machine upgrade steps

jsingle:
@zacharym these steps look right to you?  Also is this the right place to put them?


thieman:
I think it's worth adding a link to the changelog entry for Docker Machine sending people to this. Also it doesn't necessarily need to live in our docs, it could be an external gist or something. Seems sort of out of place here, especially once we're past 0.6.0.

jsingle:
Good suggestions :+1:

jsingle:
https://gist.github.com/jsingle/bd1d8c04f45040b24c28

zacharym:
How well tested are `dusty disk backup` and `dusty disk restore`?

zacharym:
should this include anything about purging boot2docker?

jsingle:
Hmm I'll test them manually again to make sure they're still working.  I don't see any integration tests for them lol

jsingle:
Maybe there should be an optional step where we show them how to purge it

zacharym:
I mean it's not our problem and people should clean up their own machines, but it's the kind of thing that I might not think to do but might actually be wise.

jsingle:
Yeah I added a note for it at the bottom

jsingle:
Also `disk backup` and `disk restore` just worked for me (backup on boot2docker and restore on Docker Machine)

PR #563, opened 2015-09-02
always get new environment per client request

jsingle:
@paetling I think we were just doing this before because `boot2docker shellinit` wouldn't give us variables back if they're already set.  This isn't the case in docker-machine, and it's actually messing people up in some cases


paetling:
Seems fine.  What was it messing up?

jsingle:
Seems to be an issue right when people upgrade, they're stuck with boot2docker env variables.  I could imagine other scenarios when they need to be changed though

PR #564, opened 2015-09-02
cast config value to string

jsingle:
@zacharym remember how you wanted to do this at one point?  Turns out we still have to do it even after we fixed where we're writing the value.  Some people already have the value written as an int, so it needs to be fixed on the read side too :(


zacharym:
makes sense

PR #566, opened 2015-09-03
unicodify bundle names

zacharym:
@jsingle I think this should do us then. If we are just going for the "best effort" strategy. I'll add some tests

### DIFF
--- a/dusty/schemas/base_schema_class.py
+++ b/dusty/schemas/base_schema_class.py
@@ -88,7 +88,7 @@ def get_specs_from_path(specs_path):
         schema = _get_respective_schema(key)
         key_path = os.path.join(specs_path, key)
         for spec_path in glob.glob('{}/*.yml'.format(key_path)):
-            spec_name = os.path.splitext(os.path.split(spec_path)[-1])[0]
+            spec_name = os.path.splitext(os.path.split(spec_path)[-1])[0].decode('utf-8')
             with open(spec_path, 'r') as f:
                 spec = yaml.load(f.read())
                 spec = DustySchema(schema, spec, spec_name, key)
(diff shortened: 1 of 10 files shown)


dusty/schemas/base_schema_class.py
```diff
@@ -88,7 +88,7 @@ def get_specs_from_path(specs_path):
         schema = _get_respective_schema(key)
         key_path = os.path.join(specs_path, key)
         for spec_path in glob.glob('{}/*.yml'.format(key_path)):
-            spec_name = os.path.splitext(os.path.split(spec_path)[-1])[0]
+            spec_name = os.path.splitext(os.path.split(spec_path)[-1])[0].decode('utf-8')
             with open(spec_path, 'r') as f:
                 spec = yaml.load(f.read())
```
jsingle: did you check that this returns unicode?

jsingle:
Not sure what the last you and travis decided about tests is.  But this PR should have at least a couple tests in it that show things working end-to-end

zacharym:
yeah I'm with you. One thing at a time

thieman:
tests or don't merge it. Also I don't understand why we are doing this a a bunch of tiny one off PRs that will be hard to back out and understand as a single unit.

zacharym:
wahh, this fails because compose calls `pyyaml.safe_load` :/ which can't handle unicode

zacharym:
@jsingle this takes care of the error propagation to the client

zacharym:
Is there any reason that we shouldn't change every single byte string to be a unicode string? @thieman any thoughts there?

jsingle:
should we be sprinkling unicode characters into our existing fixtures?  Right now we only know that the one bundles command supports it

zacharym:
yep, that's the plan

jsingle:
ah sweet :+1:

zacharym:
@jsingle the thing that is conspicuously missing here is anything that would require passing a unicode character to `self.run_command` in an actual test. I need to modify the base class `.run_command` in order to fix that.

PR #567, opened 2015-09-04
Js git known hosts

### LINKED ISSUE #554: Trouble with /private/var/root/.ssh/known_hosts when using ssh:// repos
When i run `dusty up ` with a repo defined as `ssh://git@git.doman.com:7999/path/to/repo` I get this error message `ERROR: 'Error when fetching: fatal: Could not read from remote repository.' returned with exit code 2`

However, if I go to `/etc/dusty/repos/path/to/repo` and run `sudo git pull` I get the `The authenticity of host '[git.domain.com]:7999 ([xx.xx.xx.xx]:7999)' can't be established.... Are you sure you want to continue connecting (yes/no)?` question. If I confirm the host is added to `/private/var/root/.ssh/known_hosts` and `dusty up` works just fine. Unfortunately the known_hosts entry disappears when I restart my mac. (and it's not exactly ideal that I have to go through this procedure every time I restart) 

Can this be fixed any way, so that dusty's git pull uses my known_hosts file located in `$HOME/.ssh/known_hosts`?

jsingle:
@paetling 

Resolves #554

### DIFF
--- a/dusty/commands/repos.py
+++ b/dusty/commands/repos.py
@@ -1,4 +1,5 @@
 import os
+import urlparse
 
 from prettytable import PrettyTable
 
@@ -9,6 +10,7 @@ from ..log import log_to_client
 from .. import constants
 from ..payload import daemon_command
 from ..parallel import parallel_task_queue
+from ..systems.known_hosts import ensure_known_hosts
 
 @daemon_command
 def list_repos():
@@ -41,7 +43,6 @@ def _manage_repo(repo_name):
     else:
         log_to_client('No overriden repos found by name {}'.format(repo_name))
 
-
 @daemon_command
 def manage_repo(repo_name):
     _manage_repo(repo_name)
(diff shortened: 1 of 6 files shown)


dusty/commands/repos.py
```diff
@@ -69,9 +81,12 @@ def update_managed_repos(force=False):
     specs_repo = get_specs_repo()
     if not specs_repo.is_overridden:
         log_to_client('Updating managed copy of specs-repo before loading specs')
+        add_known_hosts_for_repos([specs_repo])
```
paetling: not sure this is the right time to do this.

paetling: this method is for updating manged_repos not adding stuff to ssh known_hosts

dusty/preflight.py
```diff
@@ -106,7 +94,6 @@ def preflight_check():
     _ensure_run_dir_exists()
     _ensure_config_dir_exists()
     _ensure_command_files_dir_exists()
-    _ensure_github_known_host()
```
paetling: we cannot do it here because of dynamic repo urls?

dusty/systems/known_hosts/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+        for host in hosts:
+            if host not in contents:
+                logging.info('Adding {} ssh key to roots ssh known_hosts file'.format(host))
+                command = ['sh', '-c', 'ssh-keyscan -t rsa {}'.format(host)]
+                result = check_output(command, demote=False)
+                contents += result
+                modified = True
+        if modified:
+            f.seek(0)
+            f.write(contents)
```
paetling: why not just append to the end?

dusty/preflight.py
jsingle: preflight has no knowledge of a user's specs

dusty/systems/known_hosts/__init__.py
jsingle: We could append to the end, and keep track of the previous contents and the added contents in two separate variables.  Or we could do this

dusty/commands/repos.py
jsingle: We need to add to known hosts before updating managed repos...

paetling: still not the right place to do it.  Could do it in up before we call update_managed_repos?

jsingle: Not sure I agree... there's also the cli `dusty repos update`.  Also we need to do it in two steps since we don't know the hosts in the specs until we pull the specs repo

PR #572, opened 2015-09-08
add assets to specs documentation

jsingle:
@thieman

### DIFF
--- a/docs/specs/app-specs.md
+++ b/docs/specs/app-specs.md
@@ -38,6 +38,23 @@ mount: /my-app
 
 All Dusty commands (once and always) are started in the `mount` directory.  You might need to change to or copy from another directory.
 
+## assets
+
+```
+assets:
+  - name: GITHUB_KEY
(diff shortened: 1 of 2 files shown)


docs/specs/app-specs.md
```diff
@@ -38,6 +38,23 @@ mount: /my-app
...
+  - name: GITHUB_KEY
+    path: /root/.ssh/id_rsa
+  - name: AWS_KEY
+    path: /root/.aws_key
+    required: false
+```
+
+Assets are files which your containers need access to, but which you don't have in a repository.  The main usecase here is for private keys.
+
+You must register a local file as an asset with the `dusty assets register` command. When you register the file, its contents are copied to your VM and shared with your running Dusty containers. Assets are placed in your container at the absolute path specified at `path`.
```
thieman: It's `set`, isn't it?

jsingle: oh yeah lol

PR #575, opened 2015-09-08
better assets fix for non-running VM

jsingle:
@thieman makes all assets commands handle lack of started VM gracefully


dusty/commands/assets.py
```diff
@@ -18,6 +21,8 @@ def _get_string_of_set(items):
     return ', '.join(sorted(items))
 
 def list_all():
+    initialize_docker_vm()
+    log_to_client('Listing assets used by active apps and libs')
```
thieman: Do we need this?

PR #579, opened 2015-09-09
Js url from containers

### LINKED ISSUE #555: Allow containers to access Dusty Specified Urls
Since Dusty is now running nginx in a container, it should now be pretty easy to allow all containers to access these urls.  The best way to do this is probably to add entries into all containers `/etc/hosts` file that redirects that url out of the container to the vm.  This would then hit the nginx exposed port and go through nginxs routing layer to the correct container.

jsingle:
@paetling 

Resolves #555

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -2,6 +2,9 @@
 
 ## 0.6.2 (In Progress)
 
+* **New**
+    * Host names specified in app specs (`host_forwarding.host_name`) can now be used from inside any Dusty app container - previously they only worked from the host Mac.
+
 
 ## 0.6.1 (September 9, 2015)
 
--- a/dusty/command_file.py
+++ b/dusty/command_file.py
@@ -39,8 +39,17 @@ def _tee_output_commands(command_to_tee):
     ]
     return commands
 
-def _get_once_commands(app_spec):
-    once_commands = app_spec['commands']['once']
+def _hosts_export_commands(port_spec):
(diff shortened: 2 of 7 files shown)


dusty/command_file.py
```diff
@@ -39,8 +39,17 @@ def _tee_output_commands(command_to_tee):
     ]
     return commands
 
-def _get_once_commands(app_spec):
-    once_commands = app_spec['commands']['once']
+def _hosts_export_commands(port_spec):
+    if not port_spec['hosts_file']:
+        return []
+    commands = ['DOCKERHOST=`/sbin/ip route|awk \'/default/ { print $3 }\'`']
```
paetling: isn't this just a fixed ip defined by docker?

jsingle: I don't think so.  The interwebs as well as our codebase told me to do this

paetling:
ok. seems legit.  this is pretty awesome

jsingle:
Yeah it was real easy to implement, you were right

docs/changelog.md
```diff
@@ -2,6 +2,9 @@
 
 ## 0.6.2 (In Progress)
 
+* **New**
+    * Host names specified in app specs (`host_forwarding.host_name`) can now be used from inside any Dusty app container - previously they only worked from the host Mac.
```
thieman: Grammar

PR #580, opened 2015-09-10
apply natdnshostresolver fix before vm starts

jsingle:
@thieman this seems like the most unobtrusive way to do this.  We can't do it right after `docker-machine create`, since that command actually starts the VM as well.  I also can't find a `VBoxManage` command that will show us the current value of the setting, so it's not something that we can detect and apply as needed

### DIFF
--- a/dusty/systems/virtualbox/__init__.py
+++ b/dusty/systems/virtualbox/__init__.py
@@ -54,6 +54,17 @@ def _dusty_vm_exists():
             return True
     return False
 
+def _apply_nat_dns_host_resolver():
+    """
+    This will make the Dusty VM always use the host's DNS resolver for lookups.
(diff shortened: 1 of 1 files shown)


dusty/systems/virtualbox/__init__.py
```diff
@@ -54,6 +54,11 @@ def _dusty_vm_exists():
             return True
     return False
 
+def _apply_nat_dns_host_resolver():
+    check_and_log_output_and_error_demoted(
```
thieman: Please add a docstring here explaining what this is doing and why we need it

thieman:
Seems legit, this kind of thing is way more convenient now that we can assume we own the entire VM

PR #581, opened 2015-09-11
Use less greedy 10.174.249.x subnet for NAT routing

### LINKED ISSUE #573: VBox claims 10.0.2.x for itself by default
VBox uses 10.0.2.x for NAT routing by default. This is pretty damn greedy, since I think 10.0.x.x is a fairly commonly used subnet.

Let's alter the default VM config to use a less-common subnet. Something like this should work:

```
VBoxManage modifyvm dusty --natnet1 "10.174.254/24"
```

thieman:
@jsingle Closes #573.


jsingle:
lgtm :zap:

PR #582, opened 2015-09-11
Add initial socket ack to prevent false "running previous commands" message

thieman:
@jsingle

### DIFF
--- a/dusty/cli/__init__.py
+++ b/dusty/cli/__init__.py
@@ -80,7 +80,7 @@ def _run_command(sock, command):
         data = sock.recv(65535)
         timer.cancel()
         if data:
-            stripped = data.decode('utf-8').replace(constants.SOCKET_TERMINATOR, '').replace(constants.SOCKET_ERROR_TERMINATOR, '')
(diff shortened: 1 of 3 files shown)


dusty/cli/__init__.py
```diff
@@ -82,7 +82,9 @@ def _run_command(sock, command):
         if data:
             stripped = data.decode('utf-8').replace(constants.SOCKET_TERMINATOR, '').replace(constants.SOCKET_ERROR_TERMINATOR, '')
             sys.stdout.write(stripped)
-            if data.endswith(constants.SOCKET_TERMINATOR):
+            if stripped == constants.SOCKET_ACK:
```
jsingle: do we not want `startswith` here?

jsingle:
Looks good if you're sure we don't want `startswith` instead of ==, and it works in practice.  This is great, I  hate this bug

dusty/cli/__init__.py
thieman: Should be equivalent, I think. The entire send is just the socket ack.

jsingle: So `recv` reads only 1 `sendall`'s worth at once?

thieman: I think the complete answer there is a little complicated but with this tiny data size I am seeing that that's how it's working in practice

thieman:
Alright actually I did end up needing to strip it out, it gets grouped with other messages sometimes

PR #586, opened 2015-09-14
Cut 0.6.2

thieman:
@jsingle Am I missing anything?


jsingle:
Hmm don't think so.  I think our release checklist in our documentation is up to date

PR #591, opened 2015-09-30
Dependency fixes for integration tests

thieman:
@jsingle Also makes the usage of `ip` for Dockerhost discovery more flexible (will search through the PATH instead of always looking in /sbin)


jsingle:
As long as tests pass...

PR #601, opened 2015-11-23
Unhardcode Docker bridge IP and discover it at runtime

### LINKED ISSUE #599: Host forwarding broken with Docker 1.9
Relevant change referenced in https://github.com/docker/docker/issues/17744. Basically, the Docker daemon defaulted to creating the internal bridge on `172.17.42.1` prior to Docker 1.9. We optimistically hard-coded this as a constant in Dusty under the assumption that this would work for most people.

They changed this default as of Docker 1.9, so we'll need to be a bit more intelligent about determining the VM's IP on the bridge going forward.

In the meantime, it is possible to work around this with a Docker 1.9 install by adding the following to `EXTRA_ARGS` inside `/var/lib/boot2docker/profile` in your VM:

```
--bip 172.17.42.1/24
```

thieman:
@jsingle Fixes #599


dusty/compiler/nginx/__init__.py
```diff
@@ -27,32 +27,32 @@ def _nginx_server_name_string(port_spec):
...
     """This will output the nginx stream config string for specific port spec """
     server_string_spec = "\t server {\n"
     server_string_spec += "\t \t {}\n".format(_nginx_listen_string(port_spec))
-    server_string_spec += "\t \t {}\n".format(_nginx_proxy_string(port_spec))
+    server_string_spec += "\t \t {}\n".format(_nginx_proxy_string(port_spec, bridge_ip))
     server_string_spec += "\t }\n"
     return server_string_spec
 
-def get_nginx_configuration_spec(port_spec_dict):
+def get_nginx_configuration_spec(port_spec_dict, docker_bridge_ip):
```
thieman: I end up having to actually get the IP in `commands/run` and flow it through all of these in order to avoid importing a system from a compiler. There lies circular dependency madness.

dusty/systems/virtualbox/__init__.py
```diff
@@ -117,6 +117,11 @@ def initialize_docker_vm():
 def get_docker_vm_ip():
     return check_output_demoted(['docker-machine', 'ip', constants.VM_MACHINE_NAME]).rstrip()
 
+@memoized
+def get_docker_bridge_ip():
```
jsingle: since it's memoized, you could just call the function instead of passing it all over the place as an argument

jsingle: Might be simpler

dusty/compiler/nginx/__init__.py
jsingle: right ok

jsingle:
changelog reminder but :shipit:

dusty/systems/virtualbox/__init__.py
thieman: see note about circular dependencies

PR #602, opened 2015-11-24
Update SSH_AUTH_SOCK when it changes

thieman:
@paetling cc @tleach

### DIFF
--- a/dusty/config.py
+++ b/dusty/config.py
@@ -111,8 +111,9 @@ def _load_ssh_auth_pre_yosemite():
 def _set_ssh_auth_sock(ssh_auth_sock):
     daemon_warnings.clear_namespace('ssh')
     if ssh_auth_sock:
-        logging.info("Setting SSH_AUTH_SOCK to {}".format(ssh_auth_sock))
-        os.environ['SSH_AUTH_SOCK'] = ssh_auth_sock
+        if ssh_auth_sock != os.getenv('SSH_AUTH_SOCK'):
+            logging.info("Setting SSH_AUTH_SOCK to {}".format(ssh_auth_sock))
+            os.environ['SSH_AUTH_SOCK'] = ssh_auth_sock
     else:
         daemon_warnings.warn('ssh', 'SSH_AUTH_SOCK not determined; git operations may fail')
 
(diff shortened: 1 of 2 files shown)


dusty/config.py
```diff
@@ -111,8 +111,9 @@ def _load_ssh_auth_pre_yosemite():
 def _set_ssh_auth_sock(ssh_auth_sock):
     daemon_warnings.clear_namespace('ssh')
     if ssh_auth_sock:
-        logging.info("Setting SSH_AUTH_SOCK to {}".format(ssh_auth_sock))
-        os.environ['SSH_AUTH_SOCK'] = ssh_auth_sock
+        if ssh_auth_sock != os.getenv('SSH_AUTH_SOCK'):
+            logging.info("Setting SSH_AUTH_SOCK to {}".format(ssh_auth_sock))
+            os.environ['SSH_AUTH_SOCK'] = ssh_auth_sock
```
paetling: what is calling ssh_auth_sock right now?

paetling: how is this code any different than before?   Before you set it every time, now you are only setting it if they are different.   But if they are the same, setting the value does not really change anything?

thieman: This path gets called on every command. Before I was not setting it every time, if it had been set once the code did nothing.

dusty/config.py
```diff
@@ -122,9 +123,6 @@ def check_and_load_ssh_auth():
     SSH_AUTH_SOCK environment variable to the current environment.  This allows git clones
     to behave the same for the daemon as they do for the user
     """
-    if os.getenv('SSH_AUTH_SOCK'):
-        logging.info('SSH_AUTH_SOCK already set')
-        return _set_ssh_auth_sock(os.getenv('SSH_AUTH_SOCK'))
```
thieman: This block is the entrypoint, not the stuff up there. This deleted code was preventing it from doing anything if the envvar had ever been set.

thieman: Sorry this is confusing because of the launchctl stuff, let me try again

thieman: This return statement was doing basically nothing, it could just as easily have been a bare `return`. The updated value of SSH_AUTH_SOCK does not live in the environment, you have to get it through launchctl through the functions below this return. By removing this branch, we run those functions on each command to get the true value of the variable. We then pass the *actual* updated value to `_set_ssh_auth_sock` rather than what's already in our env, which can be stale.

paetling: ahh ok.

paetling:
:ghost: :cow:

Issue #561, opened 2015-08-31
Optional branch as part of repo setting


torbjornvatn:
I would be nice to have a way to specify a git branch for the `repo` settings in app/lib specs

E.g. 

``` yaml
repo:
  url: ssh://git@git.domain.com/path/to/repo
  branch: dev
```

Any chance for that happening?

(I should probably learn get better at Python, so I could send PRs for all these feature requests)

thieman:
+1, preferably we can find a way to do this without breaking backwards spec compatibility

torbjornvatn:
Agreed, what about:

```python
repo: [ssh://git@git.domain.com/path/to/repo, dev]
```

thieman:
We probably also want to support this for the specs repo, which is different from all the other ones currently since it's set in config rather than through specs

apeace:
One way it could be done while maintaining backwards compatibility:

```yaml
repo: ssh://git@git.domain.com/path/to/repo#branchname
```

This is how it is done in NPM, perhaps other projects as well.

thieman:
I like the approach of having the value always be a string and using `#branch`. Probably not too difficult to implement.

PR #605, opened 2016-01-06
Allow repos to point to alternate branches

apeace:
This is a solution to gamechanger/dusty#561. It allows you to have a repo setting like:

```yaml
repo: https://github.com/apeace/dusty-flask-example.git#differentbranch
```

This seems to work for me locally, but I am unfamiliar with the codebase so let me know if I've missed anything. If you let me know how to run the tests, I can follow with some test coverage tomorrow.

### DIFF
--- a/dusty/source.py
+++ b/dusty/source.py
@@ -30,10 +30,16 @@ class Repo(object):
         # remote path is either of actual remote (github.com/gamechanger/dusty) or
         # other local repo (/repos/dusty) format
         self.remote_path = remote_path
+        self.branch = 'master'
+        # remote repos may end with a branch name, like github.com/gamechanger/dusty#branch
+        if not self.is_local_repo and '#' in remote_path:
+            pieces = remote_path.split('#')
+            self.remote_path = pieces[0]
+            self.branch = pieces[1]
 
     def __eq__(self, other):
         if isinstance(other, self.__class__):
-            return self.remote_path == other.remote_path
+            return self.remote_path == other.remote_path and self.branch == other.branch
         return False
 
     def __hash__(self):
@@ -145,7 +151,7 @@ class Repo(object):
     def get_latest_commit(self):
         repo = git.Repo(self.managed_path)
         for ref in repo.refs:
-            if ref.path == 'refs/remotes/origin/master':
+            if ref.path == 'refs/remotes/origin/{}'.format(self.branch):
                 return ref.commit
         return repo.remote().fetch()[0].commit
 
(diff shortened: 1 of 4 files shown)


thieman:
+testonce

dusty/source.py
```diff
@@ -158,19 +164,19 @@ def local_is_up_to_date(self):
...
-        commits from master to bring the local copy up to date."""
+        commits from the active branch to bring the local copy up to date."""
         self.ensure_local_repo()
 
         logging.info('Updating local repo {}'.format(self.remote_path))
 
         managed_repo = git.Repo(self.managed_path)
         with git_error_handling():
-            managed_repo.remote().pull('master')
+            managed_repo.remote().pull(self.branch)
```
thieman: Do we need an extra check here now to make sure this branch is checked out before pulling? I'm concerned about getting in a dirty state if the local copy is currently on master, then I change to a different branch in Dusty and hit this code path.

thieman:
We can look into the testing failure there, doesn't seem to be related to your change.

You need to add unit tests to `tests/unit/source_test.py` around specifying a branch. Ideally we should also modify one of the testing configs in `tests/unit/compiler/test_configs` to use the branch spec and update all related tests for that config to show the changes.

We will also need doc changes to `app-specs.md` and `lib-specs.md` since this changes the accepted format of the repo value.

thieman:
Also there is some guidance on running tests locally here: http://dusty.readthedocs.org/en/stable/contributing/

dusty/source.py
apeace: Makes sense, I will take a look at that scenario tomorrow

apeace:
Thanks so much! I'll work on some tests and follow up tomorrow.

apeace:
Just added a few unit tests for the branches functionality. I'm still planning on looking into the code path you commented on, and also update a config in `tests/unit/compiler/test_configs`.

thieman:
+testonce

thieman:
The unit tests in CI should pass if you rebase off the current master.

apeace:
@thieman Rebased.

I had another thought about this change. It seems like Dusty should check out different branches of the same repo to a different directory. That way, if one bundle builds two apps off of two different branches of the same repo, there will be no conflict. Will it be necessary to do this, or does the local repo only need to be on the right branch during the build step?

thieman:
> different branches of the same repo

Crap, I hadn't thought of that. That's going to introduce complexity in quite a few places, including the CLI presentation and commands.

What are your thoughts on limiting the set of defined specs to only ever reference one branch for a given repo? This would be significantly simpler, do you think it would cover most use cases?

apeace:
I do think it would cover most use-cases. Most likely, if a developer needs multiple branches of the same repo, they're going to have one of those checked out and override the repo locally. So an informative error message in this case should be good enough for a first implementation.

In theory, though, it should be possible to allow this configuration. So maybe that is another issue for another day :)

So here is my outstanding checklist for this PR:

 * Disallow managing multiple different branches of the same repo, print an informative error message
 * Take a look at the code path you commented on above
 * Update a config in `tests/unit/compiler/test_configs` to use an alternate branch

Unfortunately it may take me a couple weeks to find the time to do all this, but I will keep at it!

apeace:
@thieman I'm sorry to say I don't have enough time to finish this PR. I wish you guys the best and thanks for the great project!

PR #606, opened 2016-01-06
Ignore a broken Pylint check

thieman:
@paetling cc @apeace pylint trying to be too smart for its own good. This should fix the pylint failure in Andrew's PR.


paetling:
this stops pylint from running in this whole file or just the line below?

thieman:
I think it disables at the block level, which I believe in this case means for that function

thieman:
Remind me never to write integration tests for anything ever again. Merging.

PR #612, opened 2016-01-12
Js docker machine env fix

### LINKED ISSUE #611: Suspected compatibility issue with Toolbox 1.9.1h
Multiple users including myself are seeing this during `dusty up`

`CalledProcessError: Command '['docker-machine', 'env', 'dusty']' returned non-zero exit status 1`

Even though running this command in a local shell works just fine. We're looking into it.

jsingle:
@thieman @paetling  I think the default should be to pass the current environment to the subprocess, I'm not sure why we're not already doing that

Also for reasons I haven't investigated, `docker-machine env` now wants the `--shell` flag passed to it

Fixes #611

### DIFF
--- a/dusty/systems/docker/__init__.py
+++ b/dusty/systems/docker/__init__.py
@@ -29,7 +29,7 @@ def get_dusty_container_name(service_name):
 @memoized
 def get_docker_env():
     env = {}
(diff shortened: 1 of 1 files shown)


jsingle:
+testonce

paetling:
:ghost: :cow: conditional on tests

jsingle:
I believe the test failure will be fixed after a rebase

PR #616, opened 2016-01-13
Add Brew cask file for 0.6.5

thieman:
@jsingle cc @sporkmonger Seems to be working

### DIFF
--- a/setup/dusty.rb
+++ b/setup/dusty.rb
@@ -1,113 +1,38 @@
-class Dusty < Formula
-  desc "A local development environment manager for your Mac"
-  homepage "https://github.com/gamechanger/dusty"
-  head "https://github.com/gamechanger/dusty.git"
+cask 'dusty' do
+  version '0.6.5'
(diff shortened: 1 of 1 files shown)


setup/dusty.rb
```diff
@@ -1,113 +1,38 @@
...
-    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
-  end
-
-  test do
-    Language::Python.each_python(build) do |python, _version|
-      system python, "-c", "import dusty"
-    end
-  end
+cask 'dusty' do
+  version '0.6.5'
```
jsingle: If we need to update this each release, can you update the release checklist in our docs?

jsingle:
Sick, looks way better than the attempt to install it from source haha

setup/dusty.rb
jsingle: Or maybe there's a way to have the release job replace it automatically?

thieman: Yeah have various doc updates to make still

thieman:
Yeah after seeing your old file I definitely remember why we abandoned this path, didn't know about cask at the time (or at least I didn't)

PR #618, opened 2016-01-13
Update installation for Brew instructions

thieman:
@jsingle


jsingle:
Nice - virtualbox can be installed as a cask?

thieman:
Yessir

PR #619, opened 2016-01-14
Update installation.md

adityadalal924:
Homebrew -> Homebrew-Cask


thieman:
Would this make more sense as `Homebrew and Homebrew-Cask`? The caskroom homepage seems to assume you already have Homebrew installed.

adityadalal924:
The homepage is *technically* out of date - Homebrew now automatically taps Homebrew Cask, so a user with Homebrew doesn't need to do anything extra to get `brew cask` working - but we are still separate projects with (slightly) different goals, etc.

That being said, looking back at the wording, I actually agree that just saying Homebrew is probably more intuitive from a user's POV, so I'll go ahead and close out the PR.

PR #631, opened 2016-02-10
Raise on no bridge IP from VBox

thieman:
@paetling I just ran a `dusty up` from a powered-down VM and many of my containers failed during startup, including my internal `nginx` container. I'm still trying to track down everything that happened but one of the issues was that it couldn't grab the bridge IP, which resulted in the nginx config being malformed. I suspect this is a symptom of the `docker` daemon not being fully initialized, which may have explained the other problems as well. Going to keep digging in for other fixes.

### DIFF
--- a/dusty/systems/virtualbox/__init__.py
+++ b/dusty/systems/virtualbox/__init__.py
@@ -119,8 +119,11 @@ def get_docker_vm_ip():
 
 @memoized
 def get_docker_bridge_ip():
-    return check_output_demoted(['docker-machine', 'ssh', constants.VM_MACHINE_NAME,
-                                 "ip route | grep docker0 | awk '{print $NF}'"]).rstrip()
+    result = check_output_demoted(['docker-machine', 'ssh', constants.VM_MACHINE_NAME,
+                                   "ip route | grep docker0 | awk '{print $NF}'"]).rstrip()
+    if not result:
+        raise ValueError('Could not get Docker bridge IP from Virtualbox, VM may not be fully initialized')
+    return result
 
 def _parse_df_output(df_line):
     split_line = df_line.split()


dusty/systems/virtualbox/__init__.py
```diff
@@ -119,8 +119,11 @@ def get_docker_vm_ip():
 
 @memoized
 def get_docker_bridge_ip():
-    return check_output_demoted(['docker-machine', 'ssh', constants.VM_MACHINE_NAME,
-                                 "ip route | grep docker0 | awk '{print $NF}'"]).rstrip()
+    result = check_output_demoted(['docker-machine', 'ssh', constants.VM_MACHINE_NAME,
+                                   "ip route | grep docker0 | awk '{print $NF}'"]).rstrip()
+    if not result:
+        raise ValueError('Could not get Docker bridge IP from Virtualbox, VM may not be fully initialized')
```
paetling: should we log_to_user, so that user sees this issue, or do we catch errors somewhere and send them to the user?

thieman:
This will cause the `up` to fail immediately since nothing will catch the exception. We could maybe look into retrying if this becomes a frequent problem, but I thought an explicit failure was better than letting the `up` continue and everything be weird and broken when it finishes.

paetling:
I agree. I just want to make sure that the message gets back to the user and is not just logged in Dusty logs.  I think either way we need to raise ValueError.  I am just making sure we do not also need to log_to_client or whatever.

thieman:
The daemon sends uncaught exceptions back to the client with a special socket terminator. They are then displayed to the user.

paetling:
cool. just making sure.  :ghost: :cow:

PR #633, opened 2016-02-10
Resurrect NIC fix, Machine's defaults are also bad

thieman:
@paetling I believe this explains some of the recent complaints we've been hearing about `docker pull` performance. Under my benchmarks switching back to PCnet-FAST III gave a 50%-70% improvement in pull speeds of large layers. We got rid of this fix when we moved to Machine because we erroneously thought the problem was in using virtio, which Machine doesn't use. Turns out the problem is in using anything other than PCnet-FAST III.

Some people are trying to carry on Jordan's fight into the Machine world: https://github.com/docker/machine/issues/1942

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,11 +1,13 @@
 # Changelog
 
 ## 0.6.6 (In Progress)
+
 * **Misc**
-    * Added support of Docker Compose 1.6
+    * Added support for Docker Compose 1.6
     * `dusty disk restore` now appends the `dusty-backup` suffix to the restore path if not provided, which should make the `restore` command a bit more intuitive to use
     * Host forwarding now supports multiple forwarding rules per app to the same container port for HTTP forwarding
     * Fixed a bug where intermediate containers were not removed when creating testing images
+    * Revived a previous networking fix that was removed after Dusty moved to Docker Machine. This fix improves external networking performance by up to 50% for large file transfers.
 
 ## 0.6.5 (January 13, 2016)
 
--- a/dusty/systems/virtualbox/__init__.py
+++ b/dusty/systems/virtualbox/__init__.py
@@ -80,14 +80,15 @@ def _init_docker_vm():
         logging.info('Initializing new Dusty VM with Docker Machine')
         machine_options = ['--driver', 'virtualbox',
                            '--virtualbox-cpu-count', '-1',
-                           '--virtualbox-memory', str(get_config_value(constants.CONFIG_VM_MEM_SIZE))]
+                           '--virtualbox-memory', str(get_config_value(constants.CONFIG_VM_MEM_SIZE)),
+                           '--virtualbox-hostonly-nictype', constants.VM_NIC_TYPE]
         check_call_demoted(['docker-machine', 'create'] + machine_options + [constants.VM_MACHINE_NAME],
                            redirect_stderr=True)
 
 def _start_docker_vm():
     """Start the Dusty VM if it is not already running."""
     if not docker_vm_is_running():
-        logging.info('Starting docker-machine VM {}'.format(constants.VM_MACHINE_NAME))
+        log_to_client('Starting docker-machine VM {}'.format(constants.VM_MACHINE_NAME))
         _apply_nat_dns_host_resolver()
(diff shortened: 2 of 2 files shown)


nathanleclaire:
Very very interesting that you are seeing these results.  I'll have to do some investigating on our end.

thieman:
@nathanleclaire Do check out our original report on b2d, among other things it links to a VirtualBox bug talking about the issue. https://github.com/boot2docker/boot2docker/issues/1022#issuecomment-145849808

thieman:
We had some more chatter internally on what we were seeing. I observed that `docker pull` consistently had issues with larger images (>100MB) using the default NIC type. It would pause for long periods of time, perhaps experiencing some sort of packet loss or issue in the kernel trying to stream down a larger image. General bandwidth is the same between the two NICs, but PCnet-FAST III does not exhibit the sticking behavior.

nathanleclaire:
Do you have any sort of script to aggregate such results?  I am considering hacking together one myself to experiment but don't want to duplicate effort if you've already got one.

thieman:
I don't, I just pulled down a larger image from our private repository 5 times under each NIC type.

nathanleclaire:
OK thanks.

thieman:
I think this was mentioned in the Machine thread as well but it's also been my experience that you need multiple virtual CPUs to see any of the network degradation.

nathanleclaire:
OK, that seems a quite relevant detail.  Thanks.

paetling:
This seems good.  I am curious as to why this is issue is only showing itself when trying to load larger image layers / images?

thieman:
No idea, and since Virtualbox is both a) insanely complex and b) closed source it's unlikely we'll ever know.

nathanleclaire:
> This seems good. I am curious as to why this is issue is only showing itself when trying to load larger image layers / images?

Maybe I misunderstand the question, but @thieman asserts that the default NIC type leads to lower download bandwidth, which naturally would slow down `docker pull` due to a large amount of `docker pull` wait time being related to downloading.

paetling:
I thought the issue was with network speed not network bandwidth.  It would make more sense if it was a network bandwidth issue.  If it was a bandwidth issue, larger images, which naturally have more image layers (which i believe try to download in concurrently?), would hit some sort of bandwidth cap and download performance would degrade.  Smaller images would not see this as much because they have less image layers and thus less concurrent bandwidth needed.  

Is this what you mean @nathanleclaire or am I misunderstanding?

nathanleclaire:
> I thought the issue was with network speed not network bandwidth. It would make more sense if it was a network bandwidth issue. If it was a bandwidth issue, larger images, which naturally have more image layers (which i believe try to download in concurrently?), would hit some sort of bandwidth cap and download performance would degrade. Smaller images would not see this as much because they have less image layers and thus less concurrent bandwidth needed.

Speed of the virtualized network interface, which affects the available download bandwidth in the VM.

If your network speed is 5MB/s vs. 100MB/s, that difference is a lot more noticeable with a 1GB image than with a 2MB image.  In the latter case both connections would seem to download the image almost instantly, whereas in the former case you'll be sitting waiting for it to download a lot longer.

paetling:
I see.  So your thought is that default is slower across the board, but is only noticeable for larger images.

thieman:
Let me restate what I was seeing in my not at all scientific tests. On a test VM with:

1. Multiple vCPUs (I used 8)
2. The Machine standard NIC driver on NIC 1 (the general access one, not the host-only one)

I observed the following:

1. Overall bandwidth and latency in most cases appeared nominal.
2. During pulls which involved pulling large *layers* (the individual components of an image), I observed a tendency for the pull to stall, hang, or otherwise slow down while pulling the large layers. Performance on smaller layers seemed fine. I encountered this issue consistently on a layer that was >100MB.

This gels with the reports we've gotten from our users that pull would sometimes hang or take very long to complete. I would recommend finding or recreating an image with large layers in it to attempt to reproduce what I was seeing.

thieman:
@paetling Thoughts on merging (following rebase)?

paetling:
like it.  :ghost: :cow:

PR #634, opened 2016-02-10
Make the restore command append backup dir if not provided

### LINKED ISSUE #632: Restore command is unintuitive
It makes sense to me as a user for the `backup` and `restore` commands to take basically the same arguments. I expect this to work:

```
dusty disk backup /tmp/backup
dusty disk restore /tmp/backup
```

However, currently the correct usage requires you to point `restore` to the folder that Dusty makes inside the directory you give it (we do this to minimize the chances that we overwrite anything we shouldn't):

```
dusty disk backup /tmp/backup
dusty disk restore /tmp/backup/dusty-backup
```

thieman:
@paetling Fixes #632.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -3,6 +3,7 @@
 ## 0.6.6 (In Progress)
 * **Misc**
     * Added support of Docker Compose 1.6
+    * `dusty disk restore` now appends the `dusty-backup` suffix to the restore path if not provided, which should make the `restore` command a bit more intuitive to use
 
 ## 0.6.5 (January 13, 2016)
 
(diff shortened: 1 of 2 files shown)


paetling:
What is the reason to even still allow full specification to the directory?   Just to maintain backwards compatibility.  Seems like the syntax for backup and restore being the same is what we want, and anything else might just add confusion.

thieman:
Backwards compatibility is the main reason, plus allowing both makes the command as "do what I mean" as possible

paetling:
Ehh.  Looking at how the command works now, I hate that we make people specify the local_back_dir in the path.  The reasons i hate it are 1. inconsistency that you mention  2. We expose dusty internals to the user.   If we just implement your change we fix 1 and keep backwards compatibility, but continue to suffer from 2.  I think It allows us more flexibility and can lead to less confusion if we always look for local_back_dir for them.

paetling:
:ghost: :cow:

PR #635, opened 2016-02-10
Remove restriction on one forwarding rule per container port

### LINKED ISSUE #607: Ability to map multiple domains to same container
I have a use-case where I'd like to be able to have something like this in my app config:

```yaml
host_forwarding:
  - host_name: sub1.local.com
    host_port: 80
    container_port: 8000
  - host_name: sub2.local.com
    host_port: 80
    container_port: 8000
```

Essentially, I just want two different domains in `/etc/hosts` to map to the same container. Of course, my application is only running one process and listening on one port (8000).

I was able to get this working with the following patch:

```
diff --git a/dusty/compiler/port_spec/__init__.py b/dusty/compiler/port_spec/__init__.py
index 50218d7..0560624 100644
--- a/dusty/compiler/port_spec/__init__.py
+++ b/dusty/compiler/port_spec/__init__.py
@@ -29,8 +29,8 @@ def _add_full_addresses(host_forwarding_spec, host_full_addresses):

 def _add_container_ports(host_forwarding_spec, container_ports):
     container_port = host_forwarding_spec['container_port']
-    if container_port in container_ports:
-        raise ReusedContainerPort("{} has already been specified and used".format(container_port))
+    #if container_port in container_ports:
+    #    raise ReusedContainerPort("{} has already been specified and used".format(container_port))
     container_ports.add(container_port)

 def _add_stream_host_port(host_forwarding_spec, stream_host_ports):
```

That's not to say this is the best way to make the change--just a hack/proof of concept that demonstrates that it can be done.

thieman:
@paetling Closes #607. I think this restriction was a relic of the early builds when we ran `nginx` on the Mac side. It works perfectly fine without these checks in place now.

### DIFF
--- a/tests/unit/compiler/port_spec/init_test.py
+++ b/tests/unit/compiler/port_spec/init_test.py
@@ -3,8 +3,7 @@ from mock import patch
 from ....testcases import DustyTestCase
 from dusty.compiler.port_spec import (_docker_compose_port_spec, _nginx_port_spec,
                                       _hosts_file_port_spec, get_port_spec_document,
-                                      ReusedHostFullAddress, ReusedContainerPort,
-                                      ReusedStreamHostPort)
+                                      ReusedHostFullAddress, ReusedStreamHostPort)
 
 class TestPortSpecCompiler(DustyTestCase):
     def setUp(self):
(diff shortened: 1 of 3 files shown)


tests/unit/compiler/port_spec/init_test.py
```diff
@@ -155,25 +154,6 @@ def test_port_spec_throws_full_address_error(self):
...
-    def test_port_spec_throws_container_port(self):
-        expanded_spec = {'apps':
-                                {'gcweb':
-                                         {'host_forwarding':[{'host_name': 'local.gc.com',
-                                                              'host_port': 80,
-                                                              'container_port': 80,
-                                                              'type': 'http'},
-                                                             {'host_name': 'local.gc.com',
-                                                             'host_port': 81,
-                                                              'container_port': 80,
```
paetling: this tests if the container ports are the same.  not if the host port is the same

thieman: oh, right, need to reword things then

paetling:
Now I am confused.  In a single container, how is it possible to have two systems running on the same port and thus the need to forward to that port twice?  

I guess this does not technically speak to the nature of what is running in the container. This specification is just used to set up nginx forwarding and ensure that ports are forwarded correctly into the container. I guess with that in mind, this restriction does not need to be there, just seems a little strange to me to remove it.

thieman:
Check the linked issue, the use case is having two or more combinations on the host pointing to the same final destination in the container, e.g. `local.service.com:80` and `local.service.com:8000` and `local.service2.com:80` can all point to port 80 in the container. Nginx can handle this just fine.

PR #637, opened 2016-02-11
Add --only flag to bundles activate

thieman:
@paetling cc @tleach. QoL improvement, makes it easier to deactivate bundles en masse if you only care about running a specific bundle right now.


paetling:
:ghost: :cow:

PR #640, opened 2016-02-11
Ae retry bridge ip

paetling:
…each try   review @thieman  I was seeing integration tests fail randomly (work sometimes and then not work other times). I believe that this is the issue.


thieman:
:bear:

thieman:
@paetling merge this?

Issue #547, opened 2015-08-25
Capitalization issue with NFS and HFS


jsingle:
Currently, if you incorrectly capitalize a local path to a dusty repo, HFS on your mac will treat it fine, since HFS is case insensitive.  However NFS will fail to mount the repo if capitalization is not correct (in your `/etc/exports`).

To avoid this, just make sure your capitalization is the same as what HFS prefers.

We'll investigate building a better solution into Dusty.

thieman:
An example of this is trying to override a repo inside your `/Users` directory and forgetting to capitalize the `U`, ending with an override like `/users/travis/my-repo`.

PR #642, opened 2016-02-12
only accept paths in dusty reposs override that are accurate wrt NFS

paetling:
review @thieman  solves https://github.com/gamechanger/dusty/issues/547

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -13,6 +13,7 @@
     * Host forwarding now supports multiple forwarding rules per app to the same container port for HTTP forwarding
     * Fixed a bug where intermediate containers were not removed when creating testing images
     * Revived a previous networking fix that was removed after Dusty moved to Docker Machine. This fix improves external networking performance by up to 50% for large file transfers.
+    * Fixed a bug which allowed a directory to be written to in HFS but did not allow that same directory to be pulled using NFS.
 
 ## 0.6.5 (January 13, 2016)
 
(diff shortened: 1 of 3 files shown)


dusty/commands/repos.py
```diff
@@ -21,10 +21,19 @@ def list_repos():
...
+        if path_element not in os.listdir(recreated_path):
+            return False
+        recreated_path = "{}{}/".format(recreated_path, path_element)
+    return True
+
 @daemon_command
 def override_repo(repo_name, source_path):
     repo = Repo.resolve(get_all_repos(), repo_name)
-    if not os.path.exists(source_path):
+    if not nfs_path_exists(source_path):
```
thieman: Need a comment here explaining why we're doing this weird stuff

dusty/commands/repos.py
```diff
@@ -21,10 +21,19 @@ def list_repos():
                        repo.override_path if repo.is_overridden else ''])
     log_to_client(table.get_string(sortby='Short Name'))
 
+def nfs_path_exists(path):
```
thieman: This function seems insane if you don't know why it's here, docstring would be great

docs/changelog.md
```diff
@@ -11,6 +11,7 @@
     * Host forwarding now supports multiple forwarding rules per app to the same container port for HTTP forwarding
     * Fixed a bug where intermediate containers were not removed when creating testing images
     * Revived a previous networking fix that was removed after Dusty moved to Docker Machine. This fix improves external networking performance by up to 50% for large file transfers.
+    * Fixed a bug which allowed a directory to be written to in HFS but did not allow that same directory to be pulled using NFS.
```
thieman: Would word this more generally, just mention fixing some case-sensitivity bug or something

dusty/commands/repos.py
```diff
@@ -21,10 +21,19 @@ def list_repos():
...
+    split_path = path.lstrip('/').split('/')
+    recreated_path = '/'
+    for path_element in split_path:
+        if path_element not in os.listdir(recreated_path):
```
thieman: This does in fact return case-sensitive results, right? Any way to do a unit test with it against actual files on an HFS filesystem?

paetling: can do

PR #643, opened 2016-02-12
Optimize around some slow Machine commands

thieman:
@paetling This makes `dusty test` quite a bit faster. There are a lot more optimizations to be gained when we stop using Machine for SSH, but I'm going to do that separately.


paetling:
:ghost: :cow:  :racehorse:

PR #644, opened 2016-02-12
Optimize the hell out of SSH calls

thieman:
@paetling Running `dusty test insights all`, which runs an empty test module, takes 11.7 seconds before my changes and 4.6 seconds after them.

### DIFF
--- a/dusty/systems/virtualbox/__init__.py
+++ b/dusty/systems/virtualbox/__init__.py
@@ -14,32 +14,46 @@ from ...subprocess import check_and_log_output_and_error_demoted, check_output_d
 from ...log import log_to_client
 
 def _command_for_vm(command_list):
-    ssh_command = ['docker-machine', 'ssh', constants.VM_MACHINE_NAME]
+    ssh_command = ['ssh', '-i', _vm_key_path(), 'docker@{}'.format(get_docker_vm_ip())]
     ssh_command.extend([command_list])
     return ssh_command
 
-def _run_command_on_vm(command_list, quiet_on_success=True):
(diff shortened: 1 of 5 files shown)


dusty/systems/virtualbox/__init__.py
```diff
@@ -152,12 +166,30 @@ def initialize_docker_vm():
 
 @memoized
 def get_docker_vm_ip():
+    ssh_port = None
+    for line in _get_vm_config():
+        if line.startswith('Forwarding'):
```
paetling: what is this now?

thieman: It's a line in the massive dump that VBox gives you for the machine config. We're taking advantage of the fact that Machine (or maybe just VBox) rigs up a localhost forwarding rule for SSH to the target machine. We use this to figure out WTF the target machine's actual IP is, which we need if we want to talk to it on any port other than the one that gets forwarded for us.

dusty/systems/virtualbox/__init__.py
```diff
@@ -152,12 +166,30 @@ def initialize_docker_vm():
...
+                break
+
+    if ssh_port:
+        ip_addr = check_output_demoted(['ssh', '-i', _vm_key_path(), '-p', ssh_port,
+                                        'docker@127.0.0.1', 'ip addr show dev eth1'])
+        for line in ip_addr.splitlines():
+            line = line.strip()
+            if line.startswith('inet') and not line.startswith('inet6'):
+                ip = line.split(' ')[1].split('/')[0]
+                return ip
```
paetling: add some comments on why these work?

thieman: Oh was this not clear?

Issue #626, opened 2016-02-07
What's the prescribed way to `docker login` before `dusty up`?


mugli:
First time `dusty up` fails if it needs to pull private docker images but is not logged in to docker registry. I'm trying to document the steps for the team and confused about how to proceed without facing the error. It feels not so ok to tell everyone that *it'll fail but move on*.

I've tried `docker-machine start dusty && docker login` before `dusty up`, but before first time initialization (`dusty up`) the docker-machine isn't even ready.

Should there be a `dusty login` command that dusty will handle login without exposing the users to docker-machine? 

Or, should dusty initialize the VM and docker-machine (and also export the environment variables by `eval "$(docker-machine env dusty)"`) even before the first time `dusty up` command, maybe during installation?

Is there any other way to solve this?

thieman:
Dusty doesn't currently provide any utilities for private registries requiring login. I'll look into this a bit in the next few days and see if I can come up with a good way to bring in support, there is likely something we can do in the VM layer to make this fairly painless.

thieman:
@paetling @mugli My understanding is that `docker login` and the associated pulls of images from private registries are all controlled by the client config file, which by default is located at `~/.docker/config.json`. Here's an example of what mine looks like after I log in to the public reg:

```
{
        "auths": {
                "https://index.docker.io/v1/": {
                        "auth": "REDACTED",
                        "email": "travis.thieman@gmail.com"
                }
        },
        "HttpHeaders": {
                "User-Agent": "Docker-Client/1.10.0 (darwin)"
        }
}
```

At `pull` time, all authentication is based off of this file, the `login` command itself just populates this if it gets a successful response on the login attempt. The daemon isn't saving the login state, it just reads from this file every time you try to pull.

My current thinking is to add a `login_required` boolean key to app, service, and test specifications (anywhere you can define an image) that would mark an image as requiring login. If this is set to `True`, Dusty can check the client config at `up` time and abort if it detects any required registries do not currently have any login information. We could then present the user a fairly targeted error message telling them how to fix them problem, e.g. `Not authenticated to private registry, run docker login <registry host> to correct`. We could do this abort at the same time we abort on not having required assets, so the user doesn't have to wait long for the `up` to error out.

Does that sound like an agreeable solution?

thieman:
[...]

PR #645, opened 2016-02-15
Fix bugs with env and down machine on required login

thieman:
@paetling Fixes the bugs that @mugli identified in #626. This makes sure the VM is initialized before trying to log in, and it passes the Docker envvars to the login command. I thought this was getting done automatically, but calling the Docker binary as a subprocess is actually a new flow for us (in other instances we either use the Python client or `exec` out to replace the Dusty process). Will cut a RC after this is merged.


paetling:
:ghost: :cow:

PR #646, opened 2016-02-15
Place the once tee file outside of project root

### LINKED ISSUE #608: Tee file should not be placed in repo folder on startup
The tee file we set up as part of `once` commands is currently placed inside the `mount` directory. This is bad practice and the file should probably be placed in `/tmp` or `/var` instead.

This can contribute to certain race conditions around building images when the build context is mutated.

thieman:
@paetling Fixes #608.


paetling:
:ghost: :cow:

PR #649, opened 2016-02-15
Unbreak streaming logging on client

thieman:
@paetling Previous code broke things on the client side, this should work for both


paetling:
:ghost: :cow:

PR #652, opened 2016-02-17
Add dusty shutdown command

thieman:
@paetling cc @jcohen28 Should be fine doing this now since we maintain our own VM.

### DIFF
--- a/dusty/systems/virtualbox/__init__.py
+++ b/dusty/systems/virtualbox/__init__.py
@@ -164,6 +164,13 @@ def initialize_docker_vm():
         _ensure_cp_dir_exists()
         _ensure_assets_dir_exists()
 
+def shut_down_docker_vm():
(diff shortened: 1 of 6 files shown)


dusty/systems/virtualbox/__init__.py
```diff
@@ -164,6 +164,13 @@ def initialize_docker_vm():
         _ensure_cp_dir_exists()
         _ensure_assets_dir_exists()
 
+def shut_down_docker_vm():
+    if docker_vm_is_running():
+        check_call_demoted(['docker-machine', 'stop', constants.VM_MACHINE_NAME])
```
paetling: should we not use docker-mahine. you know, for speed.

paetling:
add to the change log

paetling:
and to docs somewhere

PR #655, opened 2016-02-26
Add awesome custom 502 page

thieman:
@paetling SO COOL! Still need to make sure it compiles the binary fine with the new package resources.


dusty/daemon.py
```diff
@@ -21,7 +22,7 @@
 from .memoize import reset_memoize_cache
 from .warnings import daemon_warnings
 from .config import refresh_config_warnings, check_and_load_ssh_auth
-from . import constants
+from . import constants, http_server
```
paetling: is this worth naming something more specific. 502_http_server?

dusty/daemon.py
```diff
@@ -116,6 +126,7 @@ def main():
     if args['--preflight-only']:
         return
     refresh_config_warnings()
+    _start_http_server()
```
paetling: do we need to tear this down on dusty down?

paetling: Or when we stop the daemon i guess?

dusty/http_server.py
```diff
@@ -0,0 +1,76 @@
...
+from collections import namedtuple
+import calendar
+import logging
+
+from flask import Flask, request, jsonify
+
+from .compiler.spec_assembler import get_assembled_specs
+from .systems.docker import get_dusty_containers, get_docker_client
+
+logging.getLogger('werkzeug').setLevel(logging.ERROR)
```
paetling: why?

dusty/http_server.py
```diff
@@ -0,0 +1,76 @@
...
+
+    app_name = _app_name_from_forwarding_info(hostname, port)
+    containers = get_dusty_containers([app_name], include_exited=True)
+    if not containers:
+        raise ValueError('No container exists for app {}'.format(app_name))
+    container = containers[0]
+
+    new_id = uuid1()
+    new_consumer = Consumer(container['Id'], datetime.utcnow())
+    _consumers[str(new_id)] = new_consumer
```
paetling: why not use the app_name as the key instead of some new random id?

dusty/http_server.py
```diff
@@ -0,0 +1,76 @@
...
+    try:
+        status = client.inspect_container(consumer.container_id)['State']['Status']
+    except Exception as e:
+        status = 'unknown'
+    new_logs = client.logs(consumer.container_id,
+                           stdout=True,
+                           stderr=True,
+                           stream=False,
+                           timestamps=False,
+                           since=calendar.timegm(consumer.offset.timetuple()))
```
paetling: if status is unknown do we still want to run this?

dusty/http_server.py
```diff
@@ -0,0 +1,76 @@
...
+
+    updated_consumer = Consumer(consumer.container_id, datetime.utcnow())
+    _consumers[str(consumer_id)] = updated_consumer
```
paetling: why not use update the offest instead of resetting a complete new object?

dusty/daemon.py
thieman: daemon settings kills it for us when main thread exits

dusty/daemon.py
thieman: /shrug

dusty/http_server.py
thieman: otherwise it logs every request at INFO level which gets really annoying since the 502 JS polls every second

dusty/http_server.py
thieman: thought stuff might get weird if you pointed two browser windows at the same page

dusty/http_server.py
thieman: good question. i'm not expecting the inspect call to fail, i mostly put the try in there in case my key lookups fail. maybe docker-py changes their API again or something. the logs command should still work as long as the container ID is legit

dusty/http_server.py
thieman: tuples are always immutable

setup/create_binaries.sh
```diff
@@ -3,8 +3,8 @@ set -e
 venv_name=venv
 rm -rf $venv_name
 virtualenv $venv_name
-$venv_name/bin/pip install "git+https://github.com/pyinstaller/pyinstaller.git@12e40471c77f588ea5be352f7219c873ddaae056#egg=pyinstaller"
+$venv_name/bin/pip install pyinstaller==3.1.1
 $venv_name/bin/pip install .
-$venv_name/bin/pyinstaller -F --runtime-hook=setup/binary-hook.py setup/bin/dusty
+$venv_name/bin/pyinstaller -F --runtime-hook=setup/binary-hook.py setup/dusty.spec
```
paetling: what do these changes do?

setup/dusty.spec
```diff
@@ -0,0 +1,35 @@
...
+             pathex=[os.path.abspath('.')],
+             binaries=None,
+             datas=[],
+             hiddenimports=[],
+             hookspath=[],
+             runtime_hooks=['setup/binary-hook.py'],
+             excludes=[],
+             win_no_prefer_redirects=False,
+             win_private_assemblies=False,
+             cipher=block_cipher)
```
paetling: What does this Analysis object do?

setup/create_binaries.sh
thieman: previously, pyinstaller generated a specs file on the fly from inspecting our source. this created a specs file very similar to the one I've committed. the only difference is that now I need to include my resource files, which pyinstaller can't do automatically (afaict).

i copied the generated specs file, added my resources files to the analysis object, then told pyinstaller to use my specs instead of trying to generate them

dusty/http_server.py
paetling: makes sensei guess

setup/dusty.spec
thieman: I think pyinstaller uses this as an entrypoint to try to figure out what to bundle into the binary

dusty/http_server.py
paetling: good point

paetling:
don't really understand the pyinstaller stuff but the rest looks fine.   :ghost: :cow:

PR #656, opened 2016-03-02
Implement Machine fix for host-only adapter

thieman:
@paetling This brings https://github.com/docker/machine/pull/3112 into Dusty. We ported the Machine code that gets the VM IP into Dusty to get around Machine's slow startup times. Unfortunately, that means we need to keep the underlying code in sync when big bug fixes go out.

As part of this, attempted to memoize the `VBoxManage showvminfo` call. Busting cache whenever we take the machine up/down to make it pick up VM config changes we sometimes make, e.g. changing the NIC. Will want to run this through its paces in RC, the memoize part might have broken stuff.

### DIFF
--- a/dusty/systems/virtualbox/__init__.py
+++ b/dusty/systems/virtualbox/__init__.py
@@ -8,7 +8,7 @@ import time
 from subprocess import CalledProcessError
 
 from ... import constants
-from ...memoize import memoized
+from ...memoize import memoized, reset_memoize_cache
 from ...config import get_config_value
 from ...subprocess import check_and_log_output_and_error_demoted, check_output_demoted, check_call_demoted, call_demoted
 from ...log import log_to_client
@@ -115,6 +115,7 @@ def _stop_docker_vm():
     """Stop the Dusty VM if it is not already stopped."""
     check_call_demoted(['docker-machine', 'stop', constants.VM_MACHINE_NAME], redirect_stderr=True)
 
+@memoized
 def _get_vm_config():
     return check_output_demoted(['VBoxManage', 'showvminfo', '--machinereadable', constants.VM_MACHINE_NAME]).splitlines()
 
@@ -157,6 +158,7 @@ def regenerate_docker_vm_certificates():
     check_call_demoted(['docker-machine', 'regenerate-certs', '-f', constants.VM_MACHINE_NAME])
 
 def ensure_docker_vm_is_started():
+    reset_memoize_cache()
     _init_docker_vm()
     # Switch VM to use PCnet-FAST III which we have observed to
     # have much better performance than the default.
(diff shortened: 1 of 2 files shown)


dusty/systems/virtualbox/__init__.py
```diff
@@ -115,6 +115,7 @@ def _stop_docker_vm():
     """Stop the Dusty VM if it is not already stopped."""
     check_call_demoted(['docker-machine', 'stop', constants.VM_MACHINE_NAME], redirect_stderr=True)
 
+@memoized
 def _get_vm_config():
     return check_output_demoted(['VBoxManage', 'showvminfo', '--machinereadable', constants.VM_MACHINE_NAME]).splitlines()
```
paetling: why is this worth memoizing?  Is it really slow?  seems like needless speedup with costs.

dusty/systems/virtualbox/__init__.py
```diff
@@ -199,29 +200,55 @@ def get_docker_vm_ip():
...
-            if line.startswith('inet') and not line.startswith('inet6'):
-                ip = line.split(' ')[1].split('/')[0]
-                return ip
-
-    # If all our crazy hacks up there don't work, we can still get the IP
-    # from Machine directly. The reason we don't do this normally is
-    # because it's really slow, taking around 600ms on my fast machine.
-    logging.warning('Could not get IP the fast way, falling back to Docker Machine')
-    return check_output_demoted(['docker-machine', 'ip', constants.VM_MACHINE_NAME]).rstrip()
+                return host_port
```
paetling: so the first one is always the one doing the local forwarding?

dusty/systems/virtualbox/__init__.py
```diff
@@ -199,29 +200,55 @@ def get_docker_vm_ip():
...
+    raise ValueError('Could not determine localhost port for SSH forwarding')
+
+def _get_host_only_mac_address():
+    # Get the number of the host-only adapter
+    vm_config = _get_vm_config()
+    for line in vm_config:
+        if line.startswith('hostonlyadapter'):
+            adapter_number = int(line[15:16])
```
paetling: is there a better way to do this, [-2:]?

dusty/systems/virtualbox/__init__.py
```diff
@@ -199,29 +200,55 @@ def get_docker_vm_ip():
...
+
+def _get_host_only_ip():
+    """Determine the host-only IP of the Dusty VM through Virtualbox and SSH
+    directly, bypassing Docker Machine. We do this because Docker Machine is
+    much slower, taking about 600ms total. We are basically doing the same
+    flow Docker Machine does in its own code."""
+    mac = _get_host_only_mac_address()
+    ip_addr = check_output_demoted(['ssh', '-o', 'StrictHostKeyChecking=no',
+                                    '-o', 'UserKnownHostsFile=/dev/null',
+                                    '-i', _vm_key_path(), '-p', _get_localhost_ssh_port(),
```
paetling: so we ask docker, which is listening on localhost at a port, what its ip address is?

dusty/systems/virtualbox/__init__.py
```diff
@@ -199,29 +200,55 @@ def get_docker_vm_ip():
...
+        raise ValueError('No host-only adapter is defined for the Dusty VM')
+
+    for line in vm_config:
+        if line.startswith('macaddress{}'.format(adapter_number)):
+            return line.split('=')[1].strip('"').lower()
+    raise ValueError('Could not find MAC address for adapter number {}'.format(adapter_number))
+
+def _ip_for_mac_from_ip_addr(ip_addr, target_mac):
+    return_next_ip = False
+    for line in ip_addr.splitlines():
```
paetling: why does ip_addr have multiple lines?

dusty/systems/virtualbox/__init__.py
```diff
@@ -199,29 +200,55 @@ def get_docker_vm_ip():
...
+        line = line.strip()
+        if line.startswith('link/ether'):
+            line_mac = line.split(' ')[1].replace(':', '')
+            if line_mac == target_mac:
+                return_next_ip = True
+        elif return_next_ip and line.startswith('inet') and not line.startswith('inet6'):
+            ip = line.split(' ')[1].split('/')[0]
+            return ip
```
paetling: generally no idea what any of this does. Seems like magic.  I see that it is similar to the stuff above but not really sure about what it is doing.

dusty/systems/virtualbox/__init__.py
paetling: I see, you reset it in ensure_docker_vm_is_started().  So should not change without caching being busted.  Still curious about speedup. We call it 2X less right?  Not like we are pinging this like crazy?

thieman: Not hitting it super hard, but it does take about 150ms each time

dusty/systems/virtualbox/__init__.py
thieman: The line looks like `hostonlyadapter2="vboxnet0"`

dusty/systems/virtualbox/__init__.py
thieman: It's the output of `ip addr show`, run it on your VM to see what it looks like

dusty/systems/virtualbox/__init__.py
thieman: there should only ever be one `Forwarding` entry in the Dusty VM

dusty/systems/virtualbox/__init__.py
thieman: No, there's a localhost port forwarded to port 22 in the VM. We can use that to SSH to the VM before we know its actual IP address.

dusty/systems/virtualbox/__init__.py
thieman: there's no easy way to ask a machine to give you the IP assigned to the device with a given MAC address. this is parsing the output of `ip addr show`, which contains both of these things, to determine one from the other

dusty/systems/virtualbox/__init__.py
paetling: worthing doing len('hostonadapater') to make it explicit?  (It might not be)

thieman: I'll go liberally sprinkle comments around once I can get integration to stop flaking out on me

dusty/systems/virtualbox/__init__.py
paetling: seems like a better variable name is needed.  network_interfaces?

dusty/systems/virtualbox/__init__.py
paetling: seems like a better variable name is needed. network_interfaces?

thieman: gonna call it `ip_addr_show` then

PR #657, opened 2016-03-03
Shut down HTTP server at upgrade to avoid immortal thread

thieman:
@jsingle Turns out `exec` and threads play about as well together as `fork` and threads do. During upgrade flow, we need to make sure to shut down the HTTP server, or we lose track of it completely (and it never dies) once we `exec` to the new binary.

### DIFF
--- a/dusty/http_server.py
+++ b/dusty/http_server.py
@@ -25,6 +25,15 @@ def _app_name_from_forwarding_info(hostname, port):
                 return app.name
     raise ValueError('Could not find app for {}:{}'.format(hostname, port))
 
+@app.route('/shutdown', methods=['POST'])
+def shutdown():
+    logging.info('Received POST request to shut down HTTP server')
+    shutdown_function = request.environ.get('werkzeug.server.shutdown')
+    if shutdown_function is None:
+        raise RuntimeError('Not running with the Werkzeug Server')
+    shutdown_function()
+    return 'Dusty HTTP server shutting down'
+
 @app.route('/register-consumer', methods=['POST'])
 def register_consumer():
     """Given a hostname and port attempting to be accessed,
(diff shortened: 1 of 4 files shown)


dusty/http_server.py
```diff
@@ -25,6 +25,15 @@ def _app_name_from_forwarding_info(hostname, port):
                 return app.name
     raise ValueError('Could not find app for {}:{}'.format(hostname, port))
 
+@app.route('/shutdown', methods=['POST'])
+def shutdown():
+    logging.info('Received POST request to shut down HTTP server')
+    shutdown_function = request.environ.get('werkzeug.server.shutdown')
```
thieman: This is a dark incantation from Flask, I don't know why Armin thought this was more straightforward than just providing an importable `shutdown` function but here we are

dusty/http_server.py
```diff
@@ -25,6 +25,15 @@ def _app_name_from_forwarding_info(hostname, port):
...
+    if shutdown_function is None:
+        raise RuntimeError('Not running with the Werkzeug Server')
```
jsingle: this should never happen right?  If the exception is raised, will the thread die anyway?

jsingle:
Is this better than having the `thread` variable accessible and stopping the server that way?

jsingle:
Hmm maybe there's not a way to just kill the thread

jsingle:
OK the flask shit is very weird but if it works :+1:

thieman:
Yeah you can't reliably kill a thread in Python, plus even if we could we need to kill it cleanly so it can relinquish its port

dusty/http_server.py
thieman: shouldn't ever happen, if it does thread won't die though

jsingle:
Cool makes sense

PR #658, opened 2016-03-04
Close file descriptors on all subprocess calls

thieman:
@paetling Subprocess works through a fork and then an exec. The fork copies the calling process, including our file descriptors (like our open port). Nothing in exec does anything to close these. This PR politely tells Python to close them for us whenever we invoke a subprocess through our helper functions. This fixes the issue where we saw VirtualBox holding on to the HTTP server's port. We forked to Docker Machine, which still had our FDs, which then forked to VBox to start the machine, then VBox still had our FDs.

I still don't totally understand why this is defaulted to False, it seems to me like in most straight-up subprocess cases (i.e. not forking to clone yourself) you'd want to close. Might research that further.


paetling:
Yeah. This makes sense.   Should have guessed this last night when I saw that manually executing the commands did not cause the issue.

PR #660, opened 2016-03-14
Increase daemon's file handles to 8192

thieman:
@jsingle Finally found what I was looking for here. I believe this should play nicely with OS X under its default settings and should stop the open file errors we've been seeing during parallelized Git operations.


jsingle:
So the default system imposed limit is less than 8192?

thieman:
256, which is ridiculous.

jsingle:
Hmm you've tested that this doesn't error out when running on a machine with default settings?

thieman:
Yeah cool Alex did some guinea piggin' for me. Going to throw out a RC and ask people here to upgrade, should know pretty much immediately if it breaks

jsingle:
OK seems legit in that case

PR #663, opened 2016-03-16
Use $http_host instead of $host in nginx config

### LINKED ISSUE #662: $http_host versus $host
Having a touch of trouble with the generated nginx config and accessing my underlying Rails app, which is running on port 3005, with the Docker host forwarded as app.local.

At the moment, the `Host` header is set to `$host`, which Rails (at least this version) interprets as `app.local:80`, rather than `app.local:3005`.  This causes redirects to be incorrectly made to `app.local`, minus the port.

Would it be feasible to change this to use `$http_host`, which includes the port information?  I've tested this locally & it works; but I'm definitely not an Nginx expert so I'm less sure how it would affect other users.

thieman:
@jsingle Fixes #662. This makes host forwarding work better when apps are forwarded on a local port other than 80. http://stackoverflow.com/questions/15414810/whats-the-difference-of-host-and-http-host-in-nginx


jsingle:
Seems legit :zap:

Issue #659, opened 2016-03-14
Services and base images from private registries


morganjbruce:
I'm having some trouble running `dusty up` in the following scenarios:

- Defining a service with an image from a private registry.
- Defining an app with `build .`, where the Dockerfile has a FROM entry referring to a container in a private registry

In both, I get the error:

```
ERROR: Get https://REGISTRY_URL/v2/CONTAINER/manifests/TAG: no basic auth credentials
ERROR: Command 'docker-compose -f /etc/dusty/compose/docker-compose.yml -p dusty up -d' returned non-zero exit status 256
```

However, running that docker-compose command by itself works without any trouble.

I've also had a read through #626 and made sure docker is logged in and that the config.json files are being updated.

Dump output here: https://gist.github.com/morganjbruce/2646e29bac94191bb251

Any suggestions on what to investigate?

thieman:
First off, thank you for a well-written submission and for providing your `dusty dump` output. :smile: 

My first thought would be whether Dusty can successfully pull the image in the private repository if you provide it as the `image` key in the spec rather than using the `build` key. Perhaps the failure or success there might point us in the right direction.

thieman:
This was the only match I could find for your issue in Compose and it's rather recent (2 weeks), you might want to try downgrading Compose a few versions to see if the issue still exists. https://github.com/docker/compose/issues/3037

morganjbruce:
Forgot to include my app/service definitions (where REPO = my private repo):

app.yml
```
repo: somewhere.git
mount: /somewhere
build: .
```

somewhere.git includes a Dockerfile that defines a base in a private repo: `FROM REPO/ruby-2.1.5:latest`

postgres.yml
```
image: REPO/postgres:origin_master
```

Running the docker-compose command with --verbose gives the following output about authentication:

```
compose.cli.verbose_proxy.proxy_callable: docker pull <- (u'REPO/postgres', tag=u'origin_master', stream=True)
docker.api.image.pull: Looking for auth config
docker.auth.auth.resolve_authconfig: Looking for auth entry for u'REPO'
docker.auth.auth.resolve_authconfig: Found u'REPO'
docker.api.image.pull: Found auth config
compose.cli.verbose_proxy.proxy_callable: docker pull -> <generator object _stream_helper at 0x10554a730>
origin_master: Pulling from postgres
```

morganjbruce:
Trying different version combinations:

[...]

PR #665, opened 2016-03-21
Demote before executing child in PTY fork

thieman:
@paetling Attempted fix for issue in #659. Believe this is a bug starting in 0.7.0.

### DIFF
--- a/dusty/commands/utils.py
+++ b/dusty/commands/utils.py
@@ -8,6 +8,9 @@ import psutil
 
 from ..systems.docker import get_docker_env
 from ..log import log_to_client, streaming_to_client
+from ..subprocess import demote_to_user
+from ..config import get_config_value
+from .. import constants
 
 def _executable_path(executable_name):
     return subprocess.check_output(['which', executable_name]).strip()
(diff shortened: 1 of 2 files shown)


dusty/commands/utils.py
```diff
@@ -35,9 +38,11 @@ def pty_fork(*args):
     updated_env.update(get_docker_env())
     args += (updated_env,)
     executable = args[0]
+    demote_fn = demote_to_user(get_config_value(constants.CONFIG_MAC_USERNAME_KEY))
 
     child_pid, pty_fd = pty.fork()
     if child_pid == 0:
+        demote_fn()
         os.execle(_executable_path(executable), *args)
```
paetling: can we not just pass this into a demoted function subprocess and keep all the demoting isolated there?

thieman: I'm pretty sure we do need to `exec` this process directly rather than fork again, which is what all the subprocess calls would do

paetling: ok.

paetling:
:ghost: :cow:

PR #676, opened 2017-02-06
Account for RepoTags is None in images return

thieman:
review @paetling cc @nrschultz


paetling:
looks good but you broke stuff

thieman:
Unrelated CI problems, unit tests passed on CI and lint is good locally. Integration tests are having some sort of separate problem, deferring fixing that until we need to do a release.

PR #678, opened 2017-05-22
/usr/local/bin not writeable by non-root (default)

cbarrett:
Homebrew sets /usr/local/bin to be writable by the user, however this is not the default permission; in that case, installing will fail. (Homebrew changes the permissions of /usr/local/bin—seems bad but ok)

Instead, write binary to a tmp file, and move it into place after escalating privs.

r? @thieman


thieman:
Seems legit. FWIW I think Homebrew stopped requiring weird permissions on /usr/local/bin a few versions ago.

CI seems sad, will have to see what's up with it

cbarrett:
> FWIW I think Homebrew stopped requiring weird permissions on /usr/local/bin a few versions ago.

Oh that's great news, thanks for the heads up 😃. And thanks for the review.

thieman:
The install script isn't directly tested by CI so whatever's up with Jenkins shouldn't block this. Mergin'.

PR #679, opened 2018-04-05
Ns upgrade python git

nrschultz:
@paetling I tested this change on @MaxMikheyenko's computer, it looks like it solves the problem we have been seeing around dusty failing to manage repositories on newer apple computers

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,10 @@
 # Changelog
 
 ## 0.7.2 (In Progress)
+* **Misc**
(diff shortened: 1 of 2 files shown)


docs/changelog.md
```diff
@@ -1,6 +1,10 @@
 # Changelog
 
 ## 0.7.2 (In Progress)
+* **Misc**
+    * Upgraded the version of GitPython that dusty is using from 1.0.1 -> 2.1.9.  This should fix an issue with how git response were being parsed, leading to dusty failing to manage its own repositories on newer Apple computers.
+    * A fix to the install script that will make it install to a temp directory, then copy over the installed binary with correct permissions to the intended installation directory.
+    * A more robust solition to parsing the `RepoTags` configuration for testing.
```
paetling: Why did you add the bottom two bullets? It seems like you are not addressing that.

PR #696, opened 2019-08-21
log the expected `/etc/exports` config

asuraiya:
@jsingle this just prints out the expected config so people can just copy paste it into their `etc/exports` file. hoping that helps a few folks.


asuraiya:
it should also only print if there is a need to update the config, which for most people shouldn't be an issue.

jsingle:
Awesome, looks great! :shipit:

